PR #36, opened 2021-03-22
Fix/collateral loan panels data

tuan-tl:
- Fix deposited amount to be calculated based on sICX ratio
- Fix borrowed amount to get from bnUSD asset
- Fix bug on load for empty account that caused app crash


src/app/components/home/LoanPanel.tsx
```diff
@@ -25,6 +26,7 @@ const LoanPanel = () => {
 
   const stakedICXAmount = useDepositedValue();
   const loanBorrowedValue = useLoanBorrowedValue();
+  const walletBalance = useWalletBalanceValue();
   const totalLoanAmount = stakedICXAmount.div(4).minus(loanBorrowedValue);
```
WeiLi512, line 30: totalLoanAmount = (stakedICXAmount / 4) - loanBorrowedValue
I think it is not correct formula.
totalLoanAmount = (stakedICXAmount /5)

src/app/components/home/LoanPanel.tsx
```diff
@@ -254,21 +183,24 @@ const LoanPanel = () => {
           </Typography>
 
           <Typography variant="p" fontWeight="bold" textAlign="center" fontSize={20}>
-            0 bnUSD
+            {formattedAmounts[Field.LEFT]} bnUSD
```
WeiLi512, line 186: I think it is not correct.
Please reference how the prototype works

src/app/components/home/LoanPanel.tsx
gedanziger: max you can borrow is 1/4 of collateral.
Max you can borrow and receive reward is 1/5 of collateral, so at 1/4 you can still borrow but not get a reward.

To get remaining amount you can borrow, take max you can borrow, and subtract already borrowed amount

WeiLi512: I thought the collateral ratio was 500%.

WeiLi512: I think we need to calculate the stakedICXAmount in USD.
stakedICXAmount * (sICX/USD price) / 4 - loanBorrowedValue

WeiLi512: If the sICX(or ICX) price drops, the borrowed amount is bigger than the total borrowable available amount. In this case, the available amount is minus. In this case, how will handle that UI?
image

gedanziger: if less than 0, you can display 0

Issue #40, opened 2021-03-25
Collateral Tab: Should say "Adjust" not "Deposit" after collateral has already been deposited


BennyOptions:
<img width="499" alt="Screen Shot 2021-03-25 at 12 03 43 AM" src="https://user-images.githubusercontent.com/43685868/112417306-ac48d180-8cfd-11eb-9945-b02128492496.png">

PR #57, opened 2021-03-25
Collateral Tab: Should say "Adjust" not "Deposit" after collateral has already been deposited #40

WeiLi512:


BennyOptions (approved):
Approve

PR #100, opened 2021-03-28
Feature line chart ratio

tuan-tl:


gedanziger:
@WeiLi512 and @tuan-tl to check tradingview library

PR #454, opened 2021-05-14
fix: sliders and text inputs for collateral and loans sections

### LINKED ISSUE #330: Fix the slider and text inputs for the collateral and loan sections
If you enter a number into loan input fields, the slider doesn't move to show the correct value, but the position details will update with the correct figures. This has caused an issue where, if you enter a number into the field, you can decrease your risk ratio by increasing your loan.

https://user-images.githubusercontent.com/63842577/115977884-00cabf80-a5d0-11eb-8a40-18588fcc2bcc.mov

It seems to fix itself if you reset it (click Cancel or adjust the collateral), but this is still a major issue because many people will try to use the text fields, and needs to be fixed before launch.

---

Related, if you enter 0 into the collateral deposited field, the slider _does_ move as far as the locked label, but the interface acts as though the collateral balance is 0:

Screen Shot 2021-04-25 at 2 11 43 PM

WeiLi512:
close #330


parrot9design (commented):
This is much more consistent now, thanks @WeiLi512. It just needs to handle for when the collateral is pushed back to zero (see [this comment](https://github.com/balancednetwork/balanced-network-interface/issues/330#issuecomment-882958486)), then it's good to ship.

BennyOptions:
@WeiLi512 please merge this PR

PR #483, opened 2021-05-29
feat: add omm markets

### LINKED ISSUE #477: Support Additional Markets
Update frontend to support adding additional markets, and support the following in testnet:

- OMM/sICX
- OMM/IUSDC
- OMM/USDb

Note that by the time they list in mainnet, other markets will occupy these IDs, so a hardcoding scheme won't work unless we can fix the id->pool map per-network. If it assists in implementation, the geometry APIs are also available.

WeiLi512:
close #477


parrot9design:
Let's update the content for these pools:

**OMM / USDS:**
> Requires an equal amount of OMM and USDS. To earn rewards from this pool, use Omm with the same wallet.

**OMM / sICX:**
> Requires an equal amount of OMM and sICX. To earn rewards from this pool, use Omm with the same wallet.


**OMM / IUSDC:**
> Requires an equal amount of OMM and IUSDC. To earn rewards from this pool, use Omm with the same wallet.

As we won’t be tracking rewards for these pools, we can also remove the `Your daily rewards` and `Total daily rewards` sections.

Screen Shot 2021-08-06 at 8 03 08 AM


----

For non-Balanced pools, we should also remove this message from the confirmation modal:

Screen Shot 2021-08-06 at 7 59 30 AM

PR #530, opened 2021-07-01
fix: add relative position to svg element

hetfly:


hetfly:
@octocat512 we can close this as well

PR #561, opened 2021-07-16
Fix: risk ratio tooltips at mobile 

Chunn:
The result after fixed
image
image


hieunguyenebox:
@parrot9design  please review

parrot9design (changes requested):
This issue is a tricky one. The layout looks fine, but it's hard to get to trigger in the first place. Most of the time when I tap the `?` on my iPhone, the tool tips briefly flicker into view, but never for long enough to see anything.

https://user-images.githubusercontent.com/63842577/126246536-60ebad40-1284-4c8d-99c5-bfd424a5e8c8.MP4



If we're able to get the tool tips to behave like the Rebalancing one, that would be ideal. I'm just not sure if the `?` placement allows for that, given that a tool tip will always appears over the top?

BennyOptions:
@tuan-tl @Chunn I spoke with @parrot9design and we decided simply to remove tooltips on mobile for now. Please remove these tooltips then ask @parrot9design to review.

Chunn:
@BennyOptions @parrot9design (cc: @hieunguyenebox ) Sorry, I'm missed your comment. It seems not only the "Risk ratio" tooltip but also the "Rebalancing" tooltip has this bug. Please check the attached video. I've fixed this bug.@parrot9design Please help me check it again. If it's not good, please tell me. I'll update based on @BennyOptions comment. However, it seems we have to remove both tooltips in SP. Thank you!

https://user-images.githubusercontent.com/17588370/126885496-0777fb64-1a2b-4bfd-95d3-c8394fd2e11d.mov

parrot9design:
Both the risk ratio and rebalancing tool tips work great now, I didn't have any difficulty with them at all (iPhone X).

They start overlapping when the screen width is below 360px, though, so can we hide the `?` and the tool tips if the screen size is smaller than this?

parrot9design:
Let's hide the Rebalancing `?` below 360px as well, then this should be good to ship.

Chunn:
@parrot9design Please check it again. Thank you!

parrot9design (approved):

parrot9design:
Approved, thanks @Chunn!

src/app/theme/index.tsx
```diff
@@ -16,14 +16,25 @@ export const sizes = {
   upSmall: 800,
   upMedium: 1000,
   upLarge: 1200,
+  smartPhone: 600,
```
WeiLi512, line 19: There is `upExtraSmall` breakpoint. remove `smartphone`.

src/app/theme/index.tsx
```diff
@@ -16,14 +16,25 @@ export const sizes = {
...
+  small: 360,
```
WeiLi512, line 20: please use the appropriate variable name.

src/app/components/Tooltip/index.tsx
```diff
@@ -1,10 +1,15 @@
 import React, { useCallback, useState } from 'react';
 
+import { isIOS } from 'react-device-detect';
 import styled from 'styled-components';
 
 import Popover, { PopoverProps } from '../Popover';
 
-const TooltipContainer = styled.div<{ wide?: boolean }>`
+const TooltipContainer = styled.div<{ wide?: boolean; small?: boolean }>`
```
WeiLi512, line 8: Use mobile-first approach.

src/app/components/QuestionHelper/index.tsx
```diff
@@ -14,9 +14,12 @@ export const QuestionWrapper = styled.div`
   outline: none;
   cursor: help;
   color: ${({ theme }) => theme.colors.text1};
+  ${({ theme }) => theme.mediaWidth.small`
```
WeiLi512, line 17: use mobile-first approach.

src/app/components/QuestionHelper/index.tsx
```diff
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 import Tooltip from 'app/components/Tooltip';
 import { ReactComponent as QuestionIcon } from 'assets/icons/question.svg';
 
-export const QuestionWrapper = styled.div`
+export const QuestionWrapper = styled.div<{ hideOnExtraSp?: boolean }>`
```
WeiLi512, line 8: `hideOnExtraSp`  just use `hidden` attribute.

WeiLi512 (changes requested):

src/app/components/Tooltip/index.tsx
Chunn: @WeiLi512 Regarding this point, It seems we only update width for risk ratio tooltips. So I think we also need to this attribute

src/app/theme/index.tsx
```diff
@@ -13,17 +13,27 @@ import texGyeAdventorRegularWoff2 from 'assets/font/tex-gyre-adventor-regular/te
...
+      ? css`
+          @media (max-width: ${sizes[label]}px) {
+            ${css(first, ...interpolations)}
+          }
+        `
+      : css`
+          @media (min-width: ${sizes[label]}px) {
+            ${css(first, ...interpolations)}
+          }
+        `;
```
Chunn, line 36: @WeiLi512 I've used this way => Because we only update width for risk ratio tooltips on SP, It seems we should not be used the mobile-first approach. Please check it again. Thank you!

src/app/theme/index.tsx
```diff
@@ -13,17 +13,27 @@ import texGyeAdventorRegularWoff2 from 'assets/font/tex-gyre-adventor-regular/te
...
-    @media (min-width: ${sizes[label]}px) {
-      ${css(first, ...interpolations)}
-    }
-  `;
+  acc[label] = (first: any, ...interpolations: any[]) =>
+    maxWidth.includes(label)
+      ? css`
+          @media (max-width: ${sizes[label]}px) {
+            ${css(first, ...interpolations)}
+          }
```
Chunn, line 30: @WeiLi512 I've used this way => Because we only update width for risk ratio tooltips on SP, It seems we should not be used the mobile-first approach. Please check it again. Thank you!

src/app/components/Tooltip/index.tsx
WeiLi512: Then just hard code the media query scope.

image

src/app/theme/index.tsx
WeiLi512: remove the code.

WeiLi512: revert the code.

WeiLi512 (approved):

WeiLi512:
Please resolve the conflict.

WeiLi512:
After BIP-2 implementation, The UI is changed. So it can occur conflicts. Please resolve the conflicts.
In BIP-2, we removed the reward threshold.
https://gov.balanced.network/t/bip2-increase-ltv-to-35/139

image

Chunn:
@WeiLi512 I've resolved the conflict. Please help me check it again and merge soon. Thank you!

parrot9design:
Hmm, those changes shouldn't have been merged like that – 12px is too low for any content in the UI, even at mobile sizes. 14px should be the absolute minimum to make sure the content is always readable.

PR #562, opened 2021-07-17
Fix: change risk ratio bar color when passing "All Collateral Locked"

Chunn:
The result after fixed:
evidence

### DIFF
--- a/src/app/components/home/PositionDetailPanel.tsx
+++ b/src/app/components/home/PositionDetailPanel.tsx
@@ -108,6 +108,8 @@ const PositionDetailPanel = () => {
 
   const isLockWarning = lockThresholdPrice.minus(ratio.ICXUSDratio).isGreaterThan(-0.01);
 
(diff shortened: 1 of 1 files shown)


parrot9design (changes requested):
This looks good, but we should make sure the space between `All collateral locked` and `Liquidated` is still done to scale: the bar shouldn't be completely filled unless you're about to be liquidated.

Chunn:
> This looks good, but we should make sure the space between `All collateral locked` and `Liquidated` is still done to scale: the bar shouldn't be completely filled unless you're about to be liquidated.

@parrot9design I've fixed it based on your comment. Please check it again. Thank you!

image

parrot9design (approved):

parrot9design:
This is behaving properly now, thanks @Chunn!

Screen Shot 2021-07-20 at 1 04 43 PM

BennyOptions:
@WeiLi512 please review

PR #575, opened 2021-07-26
feat: update sign-in modal to support iconex and hana

hetfly:


parrot9design (approved):
This works well, thanks @hetfly!

PR #580, opened 2021-07-29
Fix: centering DropdownPopper

Chunn:
**The result after fixed:**

image
image
image

### DIFF
--- a/src/app/components/CurrencyInputPanel/index.tsx
+++ b/src/app/components/CurrencyInputPanel/index.tsx
@@ -160,7 +160,7 @@ export default function CurrencyInputPanel({
           {onCurrencySelect && <StyledDropDown selected={!!currency} />}
 
           {onCurrencySelect && (
-            <PopperWithoutArrow show={open} anchorEl={ref.current} placement="bottom">
+            <PopperWithoutArrow show={open} anchorEl={ref.current} placement="bottom" offset={[0, 2]}>
               <List style={{ width: width }}>
                 <DashGrid>
                   <HeaderText>Asset</HeaderText>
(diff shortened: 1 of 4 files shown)


parrot9design (changes requested):
It looks like this change reduced the spacing between the drop-down and the content above it. Let's push it down a bit.

**Current** | **[Demo](https://balanced.network/demo/trade/)**
----------- | ----------
Screen Shot 2021-07-30 at 2 00 35 PM | Screen Shot 2021-07-30 at 2 00 47 PM

While we're at it, can we have the hover effects (highlighted text + underline) for the pool remain active while the drop-down is visible? They currently only appear on mouseover.

----

The modals are centered within the section now, which is great. But, let's make them a bit wider – they should go outside of the content around them, rather than sit flush with it, as the slider is still partially visible, but if it's wider, like in the demo, you won't see anything:

**Current** | **[Demo](https://balanced.network/demo/trade/)**
----------- | ----------
Screen Shot 2021-08-02 at 5 30 47 PM | Screen Shot 2021-08-02 at 5 32 22 PM

Chunn:
@parrot9design I've updated based on your comment.  Please check it. Thank you!

image
image

parrot9design (approved):
Much better, thanks @Chunn!

WeiLi512:
I refactored the code. Please check the code and merge it into master if you are fine. @Chunn

Chunn:
> I refactored the code. Please check the code and merge it into master if you are fine. @Chunn

@WeiLi512 It seems it's Ok. Currently, merging is blocked. Please help me merge it. Thank you!

WeiLi512 (approved):

PR #581, opened 2021-07-29
fix: 🐛 can't withdraw all collateral

hieunguyenebox:
because when a user repay, the borrowed bnusd is rounded to 2 decimal places, so there still
be very tiny remaining amount 0.000xx left in borrow

that's why you can't withdraw all collateral 

Please repay off the loan again even you see `Borrowed` = 0 to clear the tiny amount , then can withdraw all collateral


BennyOptions (approved):
it works

tuan-tl (approved):

PR #582, opened 2021-07-29
Feat: create tooltip without arrow and border

Chunn:
**The result after fixed:**

image

image

### DIFF
--- a/src/app/components/Header/index.tsx
+++ b/src/app/components/Header/index.tsx
@@ -143,7 +143,7 @@ export default React.memo(function Header(props: { title?: string; className?: s
                 Wallet
               </Typography>
               {account && (
-                <MouseoverTooltip text={isCopied ? 'Copied' : 'Copy address'} placement="left">
+                <MouseoverTooltip text={isCopied ? 'Copied' : 'Copy address'} placement="left" noArrowAndBorder>
                   <StyledAddress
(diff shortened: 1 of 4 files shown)


parrot9design (changes requested):
The tool tips should only be as wide as the text requires them to be, with 15px of padding on either side to make sure they're evenly balanced. 

To be specific, we have `padding: 10px 15px;` specified for both tool tips in the demo.

Chunn:
@parrot9design  I've updated it.  Currently, the padding value default is 12px 16px. It seems it's OK. Please check it again. Thank you!
image

parrot9design:
Hmm, I'm not sure why those are the default values, as we designed the interface with a 5px grid layout. It looks too rectangular with the current padding, so let's make it `10px 15px` (and set those values as the default). Thanks!

Chunn:
> Hmm, I'm not sure why those are the default values, as we designed the interface with a 5px grid layout. It looks too rectangular with the current padding, so let's make it `10px 15px` (and set those values as the default). Thanks!

@parrot9design I've updated the tooltip based on your comment. Please check it again. Thank you!

parrot9design (approved):
This is great, thanks @Chunn!

WeiLi512 (approved):

PR #590, opened 2021-08-04
fix: fix login modal layout

hetfly:


parrot9design (commented):
This looks much more balanced, thanks @hetfly. I noticed the `or` separator disappears when the screen width is below 600px. Is it possible to keep this around until the width is less than 350px or so?

hetfly:
I've set it to 370px, its the lowest we can go

parrot9design (approved):
That works, thanks @hetfly!

Issue #588, opened 2021-08-04
Fix minor issues for vote pages.


WeiLi512:
- [x] fix issues mentioned in this comment. https://github.com/balancednetwork/balanced-network-interface/pull/559#issuecomment-892183493
- [x] fix gap issue on safari. css gap is not working on safari.
- [x] We don't have voters count data for the first vote. so show `- voters` if there is no data.  
![image](https://user-images.githubusercontent.com/50083174/128117889-d2926f81-2ceb-4ed4-9df6-63bdb5db1305.png)

WeiLi512:
safari 14.1 adds support for flex gap

seeyijie:
14.0.2 does not support CSS gap

gitdiff75:
Safari 14.0.2 does not support flex gap

gitdiff75:
Fixed them all except for the error message for failed transaction which should be handled by the contracts:
https://github.com/balancednetwork/balanced-contracts/search?q=Balanced+tokens+need+to+be+staked

PR #591, opened 2021-08-04
Fix minor issues for vote pages. Issue #588. First proposal with endD…

gitdiff75:
…ay 68 to have '- voters'. Fixed icon size that disturb mobile view


gitdiff75:
@parrot9design All your concerns should be addressed in this commit except for the failed transaction which will be addressed in the contract. I've reached out to Dan and he will be looking into it.

parrot9design:
Thanks, @gitdiff75. 🙏 @WeiLi512, can you authorise this on Vercel so we can preview it?

parrot9design:
Looks good! Can we also add the 67% approval requirement?

Screen Shot 2021-08-04 at 8 19 12 AM

PR #596, opened 2021-08-06
feat(swap-tab): implement ratio option component

Chunn:


src/store/swap/reducer.ts
```diff
@@ -3,16 +3,26 @@ import { createReducer } from '@reduxjs/toolkit';
...
+  switchCurrencies,
+  typeInput,
+} from './actions';
 
 export interface SwapState {
   readonly independentField: Field;
   readonly typedValue: string;
   readonly [Field.INPUT]: {
     readonly currencyId: CurrencyKey | undefined;
+    readonly instantAmount?: number | undefined;
```
WeiLi512, line 21: I'd like use `percent` instead of `instantAmount`

src/store/swap/hooks.ts
```diff
@@ -255,10 +270,10 @@ export function useDerivedSwapInfo(): {
   const {
     independentField,
     typedValue,
-    [Field.INPUT]: { currencyId: inputCurrencyId },
-    [Field.OUTPUT]: { currencyId: outputCurrencyId },
+    [Field.INPUT]: { currencyId: inputCurrencyId, instantAmount: inputInstantAmount },
+    [Field.OUTPUT]: { currencyId: outputCurrencyId, instantAmount: outputInstantAmount },
   } = useSwapState();
-
+  console.log('useSwapState', useSwapState());
```
WeiLi512, line 276: remove the line.

src/app/components/trade/SwapPanel.tsx
```diff
@@ -74,19 +81,50 @@ export default function SwapPanel() {
...
 
-  const maxInputAmount = maxAmountSpend(currencyBalances[Field.INPUT]);
-  const handleMaxInput = React.useCallback(() => {
-    maxInputAmount && onUserInput(Field.INPUT, maxInputAmount.toFixed());
-  }, [maxInputAmount, onUserInput]);
+  const handleOutputInstantSelect = React.useCallback(
+    (instantAmount: number) => {
+      onInstantAmountsSelection(Field.OUTPUT, instantAmount);
+      maxOutputAmount &&
+        onUserInput(Field.OUTPUT, ((instantAmount * Number(maxOutputAmount.toFixed())) / 100).toString());
```
WeiLi512, line 124: use `maxOutputAmount.times(percent / 100).toFiexed()`

src/app/components/trade/SwapPanel.tsx
```diff
@@ -74,19 +81,50 @@ export default function SwapPanel() {
     [onUserInput],
   );
 
+  const balances = useWalletBalances();
```
WeiLi512, line 84: why do you use the `hook`? I think `currencyBalances` is enough.

src/app/components/trade/SwapPanel.tsx
```diff
@@ -74,19 +81,50 @@ export default function SwapPanel() {
...
+
   const handleInputSelect = React.useCallback(
-    (inputCurrencyKey: CurrencyKey) => onCurrencySelection(Field.INPUT, inputCurrencyKey),
-    [onCurrencySelection],
+    (inputCurrencyKey: CurrencyKey) => {
+      onCurrencySelection(Field.INPUT, inputCurrencyKey);
+      instantAmounts[Field.INPUT] &&
+        onUserInput(
+          Field.INPUT,
+          (((instantAmounts[Field.INPUT] || 0) * Number(balances[inputCurrencyKey].toFixed())) / 100).toString(),
```
WeiLi512, line 94: I think we don't need this logic.

WeiLi512 (changes requested):

parrot9design (changes requested):
- We shouldn't show the ratio options unless signed in, because there's no context. 25% of what?

Screen Shot 2021-08-10 at 10 40 38 AM

- In the `For` field, the amount to fill for each % should be calculated based on the wallet balance of the `Swap` asset. At the moment, it uses the `For` wallet balance.

Chunn:
>In the For field, the amount to fill for each % should be calculated based on the wallet balance of the Swap asset. At the moment, it uses the For wallet balance.

@parrot9design Sorry for this inconvenience I've misunderstanding. As you said it seems we only need to implement percent optional for Swap field, because It'll be calculated based on Swap asset. Do we really need percent optional for For field ?

parrot9design:
>@parrot9design Sorry for this inconvenience I've misunderstanding this point. As you said it seems we only need to implement percent options for Swap field, because For field will be calculated based on Swap asset. Do we really need percent options for For field ?

Yes, please include it in the For field, as it is in the prototype. Even though it's calculated based on your balance of the Swap asset, some people will prefer to fill in the amount they want to receive, rather than the amount they want to spend. The options only come up when you go to input an amount, so we should make both input fields behave the same.

Chunn:
@parrot9design Thanks for sharing! I've fixed based on your comment. Please help me check it again. Thank you!

src/app/components/trade/SwapPanel.tsx
```diff
@@ -87,16 +87,18 @@ export default function SwapPanel() {
...
-    [onPercentSelection, maxInputAmount],
+    [onPercentSelection, onUserInput, maxInputAmount],
   );
 
   const handleOutputPercentSelect = React.useCallback(
     (percent: number) => {
-      maxInputAmount && onPercentSelection(Field.INPUT, percent, maxInputAmount.amount.times(percent / 100).toFixed());
+      onPercentSelection(Field.OUTPUT, percent);
+      maxInputAmount && onUserInput(Field.INPUT, maxInputAmount.amount.times(percent / 100).toFixed());
     },
```
Chunn, line 100: @WeiLi512 As Parrot9 said: "In the For field, the amount to fill for each % should be calculated based on the wallet balance of the Swap asset" => It seems we need to separating setInputValue function from selectPercent function. Because we need to set percentSelected for OUTPUT field and targetValue for INPUT field. Please help me check it again. Thank you!

parrot9design (approved):
This is much better, thanks @Chunn!

WeiLi512:
> * In the `For` field, the amount to fill for each % should be calculated based on the wallet balance of the `Swap` asset. At the moment, it uses the `For` wallet balance.

I think users can be confused with the feature.

parrot9design:
> I think users can be confused with the feature.

If there's confusion, we can revisit it. Let's just get this shipped already.

WeiLi512:
I think @BennyOptions requested this feature.
Let's get feedback from @BennyOptions .
Then let's ship it.

BennyOptions:
@WeiLi512 @parrot9design @Chunn 

Just seeing this discussion now. So I'd like us to remove the percentages from the "For" section and I'll share my reasoning.

When a user is selecting the percentages (this is what I do myself) I'm thinking "I want to sell x% of this asset". The percentage number doesn't really make sense on the "For" tab, as if you click "25%", there's no basis. What would it be 25% of? The only things the user cares about are "I want to sell x% of my stack" and "if I sell x% of my stack, how much will I receive?". To answer that second question, I would still be clicking percentages based on the assets I hold in order to see how much I will receive if I sell.

Outside of all the logic I've shared, this is also how all order-book exchanges do it (I actually don't see other AMMs that offer this), and I would assume it's the same reasoning.

Chunn:
@BennyOptions Thanks for sharing! 
@BennyOptions @WeiLi512 @parrot9design I've updated it. Please help me review it again. Thank you!

BennyOptions (approved):
approve

parrot9design (approved):

tuan-tl (approved):

src/app/components/trade/SwapPanel.tsx
```diff
@@ -256,6 +258,9 @@ export default function SwapPanel() {
               onCurrencySelect={handleInputSelect}
               id="swap-currency-input"
               currencyList={CURRENCY}
+              onPercentSelect={handleInputPercentSelect}
+              percent={percents[Field.INPUT]}
+              isLoggedIn={!!account}
```
WeiLi512, line 263: You don't need to make one more field. And `isLoggedIn` is useless for the component. So remove `isLoggedIn` prop.
`onPercentSelect={account && handleInputPercentSelect}`

WeiLi512 (changes requested):

Chunn:
@WeiLi512 Updated. Please review it again. Thank you!

WeiLi512 (approved):

PR #599, opened 2021-08-09
feat: bip-2

### LINKED ISSUE #578: Remove the reward threshold from the risk ratio
In preparation for a vote to remove the reward threshold, we need to reconsider the risk ratio section. The reward threshold defined the design of this section, as it helps users optimise their rewards and manage their risk more closely. Because of it, only a small number of positions were liquidated when the market turned.

[On the forum](https://gov.balanced.network/t/lower-collateral-requirements-remove-rewards-threshold/29/), some have already voiced concern about how well they and others will manage their risk after we remove the reward threshold indicator, especially if we also lower collateral requirements. 

It may be a simple fix, like changing the label from `Reward threshold` to `Safety threshold`, but should be considered carefully from a design perspective before we submit this proposal to the community.

Related: https://github.com/balancednetwork/balanced-contracts/issues/288

WeiLi512:
close #578


parrot9design:
I think the calculation is a little bit off for the amount you can borrow? When I max out the loan, it doesn't quite hit the `All collateral locked` threshold:

Screen Shot 2021-08-09 at 2 45 45 PM

You can see that because of that, the indicator isn't red, and the full collateral balance isn't locked. 

It comes out to 288.6% collateral ratio / 34.65% LTV

It should be 285.71% collateral ratio / 35% LTV

WeiLi512:
> I think the calculation is a little bit off for the amount you can borrow? When I max out the loan, it doesn't quite hit the `All collateral locked` threshold:
> 
> Screen Shot 2021-08-09 at 2 45 45 PM
> 
> You can see that because of that, the indicator isn't red, and the full collateral balance isn't locked.
> 
> It comes out to 288.6% collateral ratio / 34.65% LTV
> 
> It should be 285.71% collateral ratio / 35% LTV

If users control loans with the slider, users can't borrow the max amount. So it will look like that.
image

parrot9design:
Okay, if I enter `0` into the Available field, this works as expected. 👍

Why can't users borrow the max amount with the slider, though? Shouldn't we only have that restriction in place for ICX collateral deposits?

parrot9design (approved):

WeiLi512:
> Okay, if I enter `0` into the Available field, this works as expected. 👍
> 
> Why can't users borrow the max amount with the slider, though? Shouldn't we only have that restriction in place for ICX collateral deposits?

many transactions of borrowing the max amount were failed. so I set the restriction.

parrot9design (changes requested):
@WeiLi512 as mentioned in Slack, let’s remove the `Safety threshold`, and reposition the `All collateral locked` label, as shown in this version of the prototype: http://balanced-development.s3-website-us-east-1.amazonaws.com/prototype-dev/

PR #608, opened 2021-08-12
fix: add more vote information

### LINKED ISSUE #595: Fix issues on the Vote page
There are a few more issues that need fixed on the Vote/Proposal pages.

Screen Shot 2021-08-06 at 7 15 38 AM

- [x] - The % voted should be the amount of eligible BALN that has already voted (with 20% being the minimum required). There's no way it's already at 99.99%. We should also cap the number to 2 decimal places.
- [x] - There's a weird visual issue at the start of the Approve and Reject bars.
- [x] - Approve needs to include `(67% required)` if it's not being taken care of in [this PR](https://github.com/balancednetwork/balanced-network-interface/pull/591).
- [ ] - We need to add links to the [discussion](https://gov.balanced.network/t/bip2-increase-ltv-to-35/139/3) and blockchain transaction.

----

Also mentioned by Dan in Slack:

> The community reported that the displayed voting weight does not include the BALN in the pools. I checked as I just voted and it shows just the staked BALN in the initial pop up, but then shows the correct full weight in the result. 

> Also, the 100% voted is not correct. It should be the sum of the for and against from the data returned by checkVote. Those are given as a percent of total voting power. For the Approve and Reject percentages they have been correctly converted to a percent of votes cast, but for the % voted number the sum can be used directly.

WeiLi512:
close #595

### DIFF
--- a/src/app/containers/ProposalPage/index.tsx
+++ b/src/app/containers/ProposalPage/index.tsx
@@ -12,6 +12,7 @@ import { Breadcrumb } from 'app/components/Breadcrumb';
 import { Button, AlertButton } from 'app/components/Button';
 import { Column } from 'app/components/Column';
 import { DefaultLayout } from 'app/components/Layout';
+import { Link } from 'app/components/Link';
 import { BoxPanel } from 'app/components/Panel';
 import { StyledSkeleton } from 'app/components/ProposalInfo';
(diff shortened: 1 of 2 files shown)


WeiLi512:
need to enhance more.

PR #612, opened 2021-08-14
fix: fix tooltip font size on mobile

hetfly:


parrot9design (changes requested):
This works, thanks @hetfly!

Let's polish up a few other pieces while we're at it: 

- Can we make the width of the tooltips (both risk ratio and rebalancing) 260px? It’s easier to read when the content is fits nicely on 2 lines.

- When the screen is below 650px, let’s make the risk ratio tooltip width 180px. You’ll need to move the `Low risk` and `Liquidated` tooltips in a little, as they’re already hard up against the edge of the screen.

- Let’s update the rebalancing tooltip content (rebalancing is automated now; it doesn’t rely on traders):

> If bnUSD is 5% above or below $1, some of your collateral may be sold and a larger amount of debt repaid to rebalance the price.

hetfly:
@parrot9design  updated,  there was already an option for wider tooltip which is 300px, is it ok like this ?

src/app/components/Tooltip/index.tsx
```diff
@@ -6,16 +6,21 @@ import styled from 'styled-components';
...
-  @media (max-width: 650px) {
-    ${props => props.small && ' width: 156px; padding: 11px;'}
-  }
-
   width: ${props => (props.wide ? '300px' : '244px')};
   padding: 10px 0.9375rem;
   line-height: 150%;
   font-weight: 400;
   font-size: 14px;
+  margin: 0 10px;
```
parrot9design, line 14: Let's remove this margin, as it makes the text look too centered.

If this was to fix the spacing issue I mentioned, you may have misunderstood what I was referring to. As soon as you reduce the screen size, the tooltips hug right up against the edge:
Screen Shot 2021-08-19 at 12 38 08 PM
Screen Shot 2021-08-19 at 12 37 55 PM

parrot9design (changes requested):
Let's shrink it to 260px if that's not too much hassle. It may seem like a small thing, but it makes the text feel more balanced: 

300px | 260px
------ | -------
Screen Shot 2021-08-19 at 12 37 18 PM | Screen Shot 2021-08-19 at 12 37 39 PM
Screen Shot 2021-08-19 at 12 37 22 PM | Screen Shot 2021-08-19 at 12 37 32 PM

src/app/components/Tooltip/index.tsx
hetfly: Tooltip size set to 260px. 
Can't do much about the screen hugging, but it never goes off the screen.

parrot9design (approved):

WeiLi512 (approved):

PR #614, opened 2021-08-15
feat: init collateral type switching

hetfly:


hetfly:
@BennyOptions  after digging around a bit, this needs lots and lots of refactoring. I'm not sure Wei would like me to do it.
@WeiLi512  this is far from working, but can you please have a look at the changes and let me know if you think this is the correct approach?

I was also thinking about changing ```depositedAmount: BigNumber``` in ```interface CollateralState``` to something like this
```
deposited: {
  'sICX': BigNumber,
  'bnUSD': BigNumber,
  ...
}
```
This would be better than just changing collateral type and reloading the values, but this is another big change that would involve another substantial refactoring.

Wei, please have a look and let me know what you think, thanks!

hetfly:
moved here https://github.com/balancednetwork/balanced-network-interface/pull/821

PR #618, opened 2021-08-18
Fix - Origination Fee

tuan-tl:


WeiLi512 (approved):
I'd like to make the constant variable if it is not critical.

BennyOptions:
Sure, we can make a constant variable for now @WeiLi512, but we should link it to blockchain API when we can. I will raise a separate issue for low priority

BennyOptions (approved):

PR #628, opened 2021-08-21
Fix: update proposal metadata

Chunn:


Chunn:
@WeiLi512 @parrot9design @BennyOptions I've updated the proposal metadata. However, there are 2 points that need to reconsider:

- StatusMap key is different from status checking. Please check attached images:
evidence1
evidence2

- We've changed "Executed" to "Enacted". It seems we also need to change "Execution Failed" to "Enaction Failed".

Could you help me check? Thank you!

parrot9design:
> We've changed "Executed" to "Enacted". It seems we also need to change "Execution Failed" to "Enaction Failed".

Let’s change it to `Failed to enact` – thanks @Chunn!

Chunn:
>Let’s change it to Failed to enact – thanks @Chunn!
@parrot9design Updated. Please check - thanks

src/app/components/ProposalStatusIcon/index.tsx
```diff
@@ -36,21 +36,35 @@ export function ProposalStatusIcon(props: ProposalStatusProps) {
...
+      const timeLeft = startTime.diff(dayjs().utc(), 'hours');
+      if (timeLeft < 0) return '';
+      return startTime.fromNow(true);
+    }
+    return '';
+  };
 
-  let endTimeStr = platformDay
-    ? dayjs()
+  const endTimeStr = () => {
```
WeiLi512, line 52: Define the time formatting functions outside of the ProposalStautsIcon component scope.

WeiLi512 (changes requested):

WeiLi512:
> @WeiLi512 How about StatusMap key? Do we need to update it to be consistent?
> image

I think there isn't `Failed Excuted` or `Failed Excution` status. 
@dbrehmer If a proposal's vote action is not executed because of an [exception](https://github.com/balancednetwork/balanced-contracts/blob/f044f18a94c43c52405b1b77b4134a7d1e710ed0/core_contracts/governance/governance.py#L243
), what status will the proposal have?

dbrehmer:
Currently we have removed the 'Failed Execution' state, but I am thinking that it may be better to handle an error in execution and set the state to 'Failed Execution'. That way it will not be buried as a failed txn that we may not see immediately. If we handle the error and set the status to 'Failed Execution' it will be easier for the frontend to see it and show that status so we know something has to be fixed.

src/utils/timeformat.ts
```diff
@@ -0,0 +1,22 @@
+import dayjs from 'dayjs';
+
+export const formatTimeStr = (targetDay, platformDay, type: 'end' | 'start') => {
```
WeiLi512, line 3: This function is difficult to understand for me. Could you refactor the code for better understanding?

src/utils/timeformat.ts
```diff
@@ -0,0 +1,22 @@
...
+  const isStart = type === 'start';
+  const targetTime = dayjs()
+    .utc()
+    .add(targetDay - platformDay - (isStart ? 0 : 1), 'day')
+    .hour(17);
+  const timeLeft = targetTime.diff(dayjs().utc(), 'hours');
+
+  if (timeLeft < 0) return '';
+  if (isStart) return targetTime.fromNow(true);
```
WeiLi512, line 12: Why does it return in this case? I think we also have to use the format of the end time in this case.
`Starting in 2 days, 2 hours`

WeiLi512 (changes requested):

src/utils/timeformat.ts
WeiLi512: @parrot9design please confirm the format in this case.

parrot9design: It should show the days and hours until voting starts/ends.

WeiLi512 (approved):

parrot9design (approved):

PR #645, opened 2021-08-26
feat: add notification to vote menu item

hetfly:


src/app/components/AppBar/index.tsx
```diff
@@ -57,7 +60,7 @@ const ListItem = styled.li`
 
 const activeClassName = 'ACTIVE';
 
-const StyledNavLink = styled(NavLink).attrs({ activeClassName })`
+const StyledNavLink = styled(NavLink).attrs({ activeClassName })<{ hasNotification?: boolean }>`
```
WeiLi512, line 63: don't change `StyledNavLink`.
please make one specific component derived from `StyledNavLink` for the vote button.

src/queries/vote/index.ts
```diff
@@ -165,3 +166,31 @@ export const useAdditionalInfoById = (id?: number) => {
...
+
+export const useHasAccountActiveProposal = () => {
+  const { account } = useIconReact();
+  const { data: platformDay } = usePlatformDayQuery();
+
+  return useQuery<boolean | undefined>(QUERY_KEYS.Vote.IsAnyProposalActive(account || ''), async () => {
+    if (account) {
+      const proposals = await bnJs.Governance.getProposals(1, 100);
+      return (
+        (await proposals
```
WeiLi512, line 178: Did you test the logic? I think it won't work. Because I think the `filter` function doesn't support an async function as a predicate.

WeiLi512 (changes requested):

src/app/components/AppBar/index.tsx
hetfly: done

src/queries/vote/index.ts
hetfly: I did and it accidentally worked :D but you were right, it wasn't functioning properly, thanks for pointing that out! 
I've moved the async part to map function.

hetfly:
@parrot9design  @octocat512 please review

parrot9design (approved):
Fantastic work, @hetfly! It's a nice touch including it on the specific proposals – will come in very handy if there are multiple votes running at once.

octocat512 (approved):

PR #683, opened 2021-09-16
Feat: add exchange warning to sICX send confirmation

Chunn:


parrot9design (approved):
Much better, thanks @Chunn!

PR #684, opened 2021-09-16
fix: rename gradient ids in stably logo

hetfly:


hetfly:
solves this https://github.com/balancednetwork/balanced-network-info/issues/53

parrot9design:
Much better, thanks @hetfly! One thing I noticed on iOS, is that the colours sometimes don't appear in the OMM and USDS logos, and there's still an issue where the BALN circle isn't visible. Any idea what's causing that to happen?

PR #689, opened 2021-09-20
feat: add sejong network

### LINKED ISSUE #670: Migrate test.balanced.network to Sejong
We need to test on Sejong testnet going forward.

@dbrehmer @gedanziger please share the contract addresses both here and on slack

octocat512:
close #670


octocat512:
The band oracle on sejong network doesn't return ICX/USD price data.

PR #697, opened 2021-09-22
Fix: update metadata on the Vote page

Chunn:


Chunn:
@parrot9design (@tuan-tl )

parrot9design:
>@parrot9design (@tuan-tl ) Regarding your comment, I've tried to reproduce this bug but I can not. Could you reproduce this bug and send to me your data?

I don't think this has anything to do with specific data/reproducing a bug, and more about the time of day you're looking at it, as it's consistently wrong for me during my morning hours. Currently (just before 7am NZT), it says the 2 votes end in `a day, 23 hours`, when it should be `2 days, 23 hours`. It usually corrects itself later in the day, but it's always wrong in the morning.

Chunn:
@parrot9design (@tuan-tl ) I've tried to reproduce in the morning. But it seems there's no unusual

- platformDay = 155
- endDay = 157

Type | Current day | End Day
------------ | ------------ | -------------
Local time | 2021-09-28 4:08:41 GTM-08 | 2021-09-29 9:08:41 GTM-08
UTC time | 2021-09-28 12:08:52 | 2021-09-29 17:08:01

=> Time Left = 1 day, 5 hours left 

Is there any problem with this?

parrot9design:
The math checks out there, but it might be worth trying between 5pm - ~2am UTC, as that’s usually when I noticed the numbers were off.

We did discover some issues with the `day.js` package, is it possible something there is causing an issue?

PR #698, opened 2021-09-23
Fix: update claiming ICX content

Chunn:


BennyOptions:
Still broken, here is the situation:

I unstaked 1 sICX

Transaction: https://tracker.icon.foundation/transaction/0x3183a92912dabbd8ec036442f55c4de53e316a917ba9984b5007a5d279074473

See the result on the UI below. It displays two activities:

1.) Unstaked 1 sICX - CORRECT
2.) Claimed 1.05 ICX - INCORRECT

Only the first transaction should be displayed. Nothing was claimed yet, so it should not display any claim transactions
Screen Shot 2021-10-05 at 11 03 56 PM

BennyOptions (approved):
UI works properly now, thanks!

src/queries/history/index.ts
```diff
@@ -36,3 +36,22 @@ export const useAllTransactionsQuery = (page: number, limit: number, account: st
     },
   );
 };
+
+export const useInternalTransactionQuery = (transaction_hash: string) => {
```
octocat512, line 40: Please use camelCase. @Chunn

octocat512 (changes requested):

PR #703, opened 2021-09-29
fix/baln allocation layout

octocat512:


parrot9design:
This looks so much better at mobile, thanks @octocat512!

Just one thing – are you able to adjust the order of the list? I’d like all the liquidity pools at the end so it doesn’t look as chaotic. It should go in this order:

- DAO fund
- Reserve
- Workers
- Borrowers
- sICX / ICX
- sICX / bnUSD
- BALN / bnUSD
- BALN / sICX
- IUSDC / bnUSD

PR #708, opened 2021-09-30
Feat/implement DAO Spending proposal

Chunn:


parrot9design:
Do we want people to be able to direct the funding towards more than one address? The `Add another asset` link was only intended for additional assets, not additional addresses.

BennyOptions:
@parrot9design that is correct.

@Chunn please remove additional address lines

Screen Shot 2021-10-01 at 12 46 28 AM

Chunn:
@parrot9design @BennyOptions Regarding this point, I've created UI based on the sample action parameter. Please check https://bicon.tracker.solidwallet.io/transaction/0x86b672e5ea7a63f569dfaee34c3bd88ffd84bdd4bd8f62f42f460040c82b4fe3
```{
    "_recipient": "hx3f01840a599da07b0f620eeae7aa9c574169a4be",
    "_amounts": [
        {
            "address": "cxae6334850f13dfd8b50f8544d5acb126bb8ef82d",
            "amount": 10000000000000000000
        },
        {
            "address": "cxc48c9c81ceef04445c961c5cc8ff056d733dfe3a",
            "amount": 10000000000000000000
        }
    ]
}
```
I'll remove the address field based @BennyOptions comment. Could you confirm for me what the data type of `_amount` will be?
Because we have only one address => It seems actions params should be
 ```{
    "_recipient": "hx3f01840a599da07b0f620eeae7aa9c574169a4be",
    "_address": "cxae6334850f13dfd8b50f8544d5acb126bb8ef82d",
    "_amounts": [
        {
            "amount": 10000000000000000000,
            "symbol": "BALN"
        },
        {
            "amount": 10000000000000000000,
            "symbol": "sICX"
        }
    ]
}
```
Please help me confirm. Thank you!

**Comment by parrot9design** [comment]
There's a bug at the moment, where if you add all possible assets and then click `Add another asset` again, you get a white screen. Once all available assets have been listed, we should remove that text link to prevent people getting into trouble.

I also noticed that once you add another asset field, you can't submit the proposal until you add a value to it. We should allow people to submit the proposal as long as at least one asset has a value specified. Any assets selected that don't have a value should be ignored when submitting.

**Comment by Chunn** [comment]
@parrot9design I've updated based on your comment. Please check again. Thank you!

**Comment by parrot9design** [comment]
> I've updated based on your comment. Please check again. Thank you!

This is looking great now, thank you!

**Review by parrot9design — APPROVED** [review]

File: src/app/components/CurrencyInputPanel/index.tsx [hunk 1]
```diff
@@ -123,6 +126,8 @@ export default function CurrencyInputPanel({
   customBalanceText,
   bg = 'bg2',
   placeholder = '0',
+  className,
+  balanceList,
```
octocat512, line 130: I'd like to make `balanceList` data type as   `[key: string]: BigNumber;`

octocat512: I will refactor the code. after that,  please review @Chunn

octocat512 (changes requested):

src/app/components/CurrencyInputPanel/index.tsx
octocat512: In the future, the bignumber data type will be changed with `CurrencyAmount` data type.

BennyOptions (approved):
approve

PR #714, opened 2021-10-05
Fix/rebalancing param

tuan-tl:


src/app/containers/NewProposalPage/constant.ts
```diff
@@ -162,7 +162,7 @@ export const PROPOSAL_CONFIG = {
       const rebalance_ratio = BalancedJs.utils
         .toLoop(Number(Object.values(ratioInputValue)))
         .div(100)
-        .toFixed();
+        .toNumber();
```
octocat512, line 165: Javascript number type can handle 16 digits integer?

tuan-tl: For this case input is percentage with 2 decimal numbers so I think it should cover enough

PR #716, opened 2021-10-06
update pool id description

rise-techiast:


parrot9design (approved):
Much better, thanks! Just thinking we should include one for the CFT pool, too, but happy for that to happen in a separate PR if that's easier.

>Requires an equal amount of CFT and sICX. To earn rewards from this pool, use [Craft](https://craft.network) with the same wallet.

PR #729, opened 2021-10-14
Feat: displaying implied ICX price

Chunn:


octocat512:
@Chunn Please write the code for others to read as possible. I think we need to define the code convention.

PR #733, opened 2021-10-17
Fix: show LedgerConfirmMessage when voting with a Ledger

Chunn:


BennyOptions:
I am trying to test this feature but there is no active proposal. I created a new vote but it was not added to the list of proposals

Transaction: https://bicon.tracker.solidwallet.io/transaction/0x4ab9d70d3e4cba5983a0db0e6be5c7575c40ca7675026c7d61f5a703a20e487d

Screen Shot 2021-10-19 at 1 10 01 AM

BennyOptions:
@rise-techiast the app does not load when I click the vercel link, it's a white screen like when the app crashes. I tried to clear cache as well, but it still never loads.

PR #736, opened 2021-10-19
Fix: Can't submit a vote even though they have the necessary staked BALN

Chunn:


rise-techiast:
> I was unable to check, but @rise-techiast if you can verify that the address provided is working properly then please merge

@BennyOptions yup, I've checked it locally and it works fine

PR #791, opened 2021-11-20
Fix: hide vote cancelled

Chunn:


BennyOptions:
@Chunn can you switch this PR to main net? @rise-techiast if Chunn is not familiar with that process please assist. Then I can see if the cancelled votes are removed as a test

BennyOptions (approved):
approved, it works

octocat512:
Thanks @Chunn 
But I'd like to filter proposals in the `useTotalProposalQuery` query. 
https://github.com/balancednetwork/balanced-network-interface/blob/d3d3f2d3ad825836c521b865e908ef37dd814710/src/queries/vote/index.ts#L138
Please add a filter function to the data variables. 

`data.filter(r => r['status'] === 'Cancelled')`

PR #793, opened 2021-11-21
feat: add custom scroll bar option

hetfly:


parrot9design:
The React customScrollbars didn't have an effect on how the scrollbars behave on Mac or iOS. We'll address this issue another way, by adding a search box to the top of the dropdown instead. Closing this for now.

PR #801, opened 2021-11-23
Refactor/remove addresses variable

### LINKED ISSUE #800: Refactor the code to not use addresses variable for getting token address
The frontend code uses the addresses constant variable to get token addresses in some parts. The addresses constant variable has to keep the only balanced related token addresses. If we add a new token, then the frontend has to add the address in the addresses variable. The front end has the tokens constant variable for token information. So use the tokens constant variable.

https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/packages/BalancedJs/addresses.ts#L3

After this refactoring, the frontend doesn't need to add a new token address to the addresses variable when adding a new token. 

- Please use address instead of using symbol. And don't use the addresses variable for getting token address. 
https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/packages/BalancedJs/contracts/IRC2.ts#L44
https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/packages/BalancedJs/contracts/IRC2.ts#L31
https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/packages/BalancedJs/index.ts#L148
- Don't use the addresses variable to get token addresses 
https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/store/pool/hooks.ts#L86
https://github.com/balancednetwork/balanced-network-interface/blob/42536c884dd905fd5b249bc4557ce6765cbffe40/src/store/pool/hooks.ts#L87

lijac:
Close #800

### DIFF
--- a/src/app/components/trade/SwapPanel.tsx
+++ b/src/app/components/trade/SwapPanel.tsx
@@ -171,13 +171,14 @@ export default function SwapPanel() {
       const minReceived = executionTrade.minimumAmountOut(new Percent(slippageTolerance, 10_000));
 
       const token = executionTrade.inputAmount.currency as Token;
+      const outputToken = executionTrade.outputAmount.currency as Token;
 
       bnJs
         .inject({ account })
         .getContract(token.address)
         .swapUsingRoute(
           BalancedJs.utils.toLoop(executionTrade.inputAmount.toExact(), currencies[Field.INPUT]?.symbol!),
-          executionTrade.outputAmount.currency.symbol!,
+          outputToken.address,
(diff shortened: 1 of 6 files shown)


octocat512:
Please remove the non-baln related token addresses in the addresses constant variable. And check if the balanced UI works correctly. @lijac

Issue #772, opened 2021-11-15
Refactor useCurrencyBalances hook in the swap panel.


octocat512:
In the swap panel, the frontend uses the wallet redux store to get the balance of a token. please check the logic here.
https://github.com/lijac/balanced-network-interface/blob/ea328fecc96bc80b43ee8110310b8a6a9bfc4852/src/store/swap/hooks.ts#L78

With the current logic, the frontend can't support user-defined custom tokens in the future. The frontend needs to get the balance of a token by address and don't have to rely on the wallet redux store. Please refactor the `useCurrencyBalances` hook in the tx routing PR if the PR is not merged.
https://github.com/lijac/balanced-network-interface/blob/ea328fecc96bc80b43ee8110310b8a6a9bfc4852/src/store/swap/hooks.ts#L78

While refactoring, Please reference the code.
https://github.com/Uniswap/interface/blob/34a58851f779e288fd874490da6fafea1b0bf2b4/src/state/wallet/hooks.ts#L107

octocat512:
https://github.com/lijac/balanced-network-interface/blob/ea328fecc96bc80b43ee8110310b8a6a9bfc4852/src/app/components/trade/LiquidityDetails.tsx#L397
Instead of using the `useWalletBalances` hook, Please use the refactored `useCurrencyBalances` hook.

Finally, remove the using `useWalletFetchBalances` hook in the TradePage component.
https://github.com/lijac/balanced-network-interface/blob/ea328fecc96bc80b43ee8110310b8a6a9bfc4852/src/app/containers/TradePage/index.tsx#L23

octocat512:
Expected Results:
To introduce uniswap-like scalability and efficiency to the codebase, useWalletbalances should be refactored.
This will help us apply future updates seamlessly

PR #807, opened 2021-11-29
feat: add custom pair

### LINKED ISSUE #539: UI/UX around listing a new pool
Other platforms allow you to list your own pool directly from the UI rather than needing to be a developer. See screenshot from Uniswap below.

Screen Shot 2021-07-06 at 5 00 04 PM

Screen Shot 2021-07-06 at 5 01 15 PM

octocat512:
close #539


octocat512:
@lijac In this PR, I need your refactoring code from https://github.com/balancednetwork/balanced-network-interface/issues/772

So instead of making the PR, please work on this PR for the 772 issue. I will close it now.

src/store/swap/hooks.ts
```diff
@@ -75,22 +77,122 @@ export function useSwapActionHandlers(): {
...
+        return agg;
+      }, ''),
+    [currencies],
+  );
+  const tokens = useMemo(
+    () =>
+      (currencies?.filter((currency): currency is Token => currency?.isToken ?? false) ?? []).filter(
+        (token: Token) => !isNativeCurrency(token),
+      ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
octocat512, line 176: Why did you disable the lint?

octocat512:
image
In the mobile, I can see the unexpected breakdown.

lijac (commented):
Review changes.

octocat512:
If we add the custom pair feature, the frontend has to consider the custom pair when calculating a route. So I'd like you to refactor the swapping logic for the custom pair feature.

- [ ] Please refactor the `useTradeExactIn` and `useTradeExactOute` hooks for the custom pair feature.
reference:
https://github.com/Uniswap/interface/blob/82843ff16aa685015ad2ab3468afb22c04770768/src/hooks/useBestV2Trade.ts#L48

- [ ] Please implement the `useV2Pairs` and `useV2Pair` hooks. Please keep the function name, parameter, and returns. The hooks will be used in the other part.
reference:
https://github.com/Uniswap/interface/blob/main/src/hooks/useV2Pairs.ts

src/app/components/SearchModal/CurrencyList.tsx
```diff
@@ -1,11 +1,51 @@
...
+import { useIconReact } from 'packages/icon-react';
 import { Flex } from 'rebass/styled-components';
 
 import CurrencyLogo from 'app/components/CurrencyLogo';
 import { ListItem, DashGrid, HeaderText, DataText, List1 } from 'app/components/List';
+import { useCurrencyBalance } from 'store/swap/hooks';
 import { Currency, Token } from 'types/balanced-sdk-core';
 
+function currencyKey(currency: Currency): string {
+  return currency.isToken ? currency.address : 'ICX';
```
lijac, line 12: `currency.isToken` is `true` for all tokens, including `ICX` (we define `ICX` as a token in `contants/tokens.ts` for convenience.).

src/store/pool/hooks.ts
```diff
@@ -355,3 +357,63 @@ export function useAPYs() {
 
   return apys;
 }
+
+export enum PairState {
```
octocat512, line 361: Please move the logic in the hooks directory like uniswap code.

src/store/pool/hooks.ts
```diff
@@ -355,3 +357,63 @@ export function useAPYs() {
...
+    return currencies.map(([currencyA, currencyB]) => [currencyA?.wrapped, currencyB?.wrapped]);
+  }, [currencies]);
+
+  useEffect(() => {
+    const fetchReserves = async () => {
+      const result = await Promise.all(
+        tokens.map(async ([tokenA, tokenB]) => {
+          if (tokenA && tokenB && tokenA.chainId === tokenB.chainId && !tokenA.equals(tokenB)) {
+            const poolId = await bnJs.Dex.getPoolId(tokenA.address, tokenB.address);
+            const reserveA = await bnJs.Dex.getPoolTotal(poolId, tokenA.address);
```
octocat512, line 382: what about using `getPoolStats` API? It can reduce API calls.

src/store/pool/hooks.ts
```diff
@@ -355,3 +357,63 @@ export function useAPYs() {
...
+        PairState.EXISTS,
+        new Pair(CurrencyAmount.fromRawAmount(token0, reserve0), CurrencyAmount.fromRawAmount(token1, reserve1)),
+      ];
+    });
+  }, [reserves, tokens]);
+}
+
+export function useV2Pair(tokenA?: Currency, tokenB?: Currency): [PairState, Pair | null] {
+  const inputs: [[Currency | undefined, Currency | undefined]] = useMemo(() => [[tokenA, tokenB]], [tokenA, tokenB]);
+  console.log('useV2Pair');
```
octocat512, line 417: Please remove console.log

src/store/pool/hooks.ts
```diff
@@ -355,3 +357,63 @@ export function useAPYs() {
...
+  NOT_EXISTS,
+  EXISTS,
+  INVALID,
+}
+
+export function useV2Pairs(currencies: [Currency | undefined, Currency | undefined][]): [PairState, Pair | null][] {
+  const [reserves, setReserves] = useState<({ reserve0: string; reserve1: string } | number | undefined)[]>([]);
+
+  const tokens = useMemo(() => {
+    setReserves(Array(currencies.length).fill(PairState.LOADING));
```
octocat512, line 372: What about moving the logic of initializing the array in the `fetchReserves` function?
The code looks strange to me.

lijac:
> image In the mobile, I can see the unexpected breakdown.

Not sure if the additional line is useful from a UX perspective. @parrot9design what do you think?

src/store/pool/hooks.ts
octocat512: image

lijac: Agreed!

src/store/pool/hooks.ts
lijac: Ok

lijac: OK

src/store/pool/hooks.ts
lijac: OK

src/store/pool/hooks.ts
lijac: Ah, yes!

parrot9design:
> > image In the mobile, I can see the unexpected breakdown.
> 
> Not sure if the additional line is useful from a UX perspective. @parrot9design what do you think?

No, definitely no value in keeping it, as there should only be a single horiziontal line. Here's the modal at desktop size:

Screen Shot 2021-12-05 at 12 56 24 PM

src/app/components/trade/LPDescription.tsx
```diff
@@ -111,18 +87,19 @@ export default function LPDescription({ baseSuplying, quoteSupplying }: ILPDescr
               Your supply
             </Typography>
             <Typography textAlign="center" variant="p">
-              {selectedPair.id !== BalancedJs.utils.POOL_IDS.sICXICX ? (
+              {!pair?.queueRate ? (
```
lijac, line 90: Using ID looks more explicit. We will keep the sICXICX pool ID constant for sure in the future, but the queueRate can be renamed or changed.

src/app/components/trade/LPDescription.tsx
```diff
@@ -131,26 +108,27 @@ export default function LPDescription({ baseSuplying, quoteSupplying }: ILPDescr
...
+          )} */}
         </Box>
         <Box width={[1, 1 / 2]}>
           <Box sx={{ margin: '15px 0 25px 0' }}>
             <Typography textAlign="center" marginBottom="5px" color="text1">
               Total supply
             </Typography>
             <Typography textAlign="center" variant="p">
-              {selectedPair.id !== BalancedJs.utils.POOL_IDS.sICXICX ? (
+              {!pair?.queueRate ? (
```
lijac, line 119: pool id vs queueRate

src/app/components/trade/SupplyLiquidityModal.tsx
```diff
@@ -267,20 +263,17 @@ export default function SupplyLiquidityModal({ isOpen, onClose, parsedAmounts, c
     changeShouldLedgerSign(false);
   };
 
-  const isQueue = selectedPair.id === BalancedJs.utils.POOL_IDS.sICXICX;
-  const pair = SUPPORTED_PAIRS.find(p => p.id === selectedPair.id);
+  const isQueue = !!pair?.queueRate;
```
lijac, line 266: > pool id vs queueRate

src/app/components/trade/SupplyLiquidityModal.tsx
```diff
@@ -186,15 +182,15 @@ export default function SupplyLiquidityModal({ isOpen, onClose, parsedAmounts, c
         .Dex.add(
           baseToken.address,
           quoteToken.address,
-          BalancedJs.utils.toLoop(pool?.baseDeposited || new BigNumber(0), selectedPair.baseCurrencyKey),
-          BalancedJs.utils.toLoop(pool?.quoteDeposited || new BigNumber(0), selectedPair.quoteCurrencyKey),
+          baseDeposit?.quotient?.toString() ?? '0',
```
lijac, line 185: Any specific reason to use `quotient` directly?

octocat512: The `BalancedJs.utils.toLoop` function will be deprecated. So I'd like you to not use the function too in the future. We will refactor that in December.

src/app/components/trade/LPDescription.tsx
octocat512: I got you

octocat512:
The Balanced Package is using the BigNumber type for handling big numbers. Instead of using BigNumber, what about using string type?

octocat512:
Deprecate `toLoop` and `toIcx` function.

octocat512:
Is there any solution to simply the LedgerConfirmMessage component?

parrot9design:
Here's the first round of feedback, based mostly on https://balanced-development.s3-website-us-east-1.amazonaws.com/prototype-dev/trade/.

## Import custom tokens

- [ ] When you click the asset fields, a modal appears. It should be contained in a dropdown instead:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 10 55 05 AM | Screen Shot 2021-12-09 at 10 57 16 AM

- [ ] Update the way new tokens are displayed in search results:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 11 11 21 AM<br>Screen Shot 2021-12-09 at 11 17 04 AM | Screen Shot 2021-12-09 at 11 11 49 AM<br>Screen Shot 2021-12-09 at 11 08 24 AM

- [ ] Tokens should be listed in descending order based on wallet balance, then alphabetically. If possible, you should also be able to click asset/wallet to change how the list is ordered.

- [ ] Update the confirmation modal:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 11 21 28 AM | Screen Shot 2021-12-09 at 11 36 07 AM

The modal for this is hidden and doesn't include the alert message, so here's the code snippet:

```
<div class="modal-content">
	<!-- Overview -->
	<h3 class="label text-center">Import asset?</h3>
	<p class="text-bold text-center">GBET</p>
	<p class="text-center">Gangster Bet</p>

	<!-- Contract address -->
	<p class="label text-center" style="width: 175px;display: block;word-wrap: break-word;margin: auto;margin-top: 5px;margin-bottom: 20px;">cxf61cd5a45dc9f91c15aa65831a30a90d59a09619</p><p class="label text-center text-red">Make sure these details are correct before you add it to Balanced on this device.</p>

	<!-- Modal actions -->
	<div class="modal-actions">
		<a href="#" class="cancel modal-close">Cancel</a> <a href="./" class="button">Import</a>
	</div>
</div>
```

- [ ] We still need to get you a default coin logo to display in the coin selector after it's been imported
Screen Shot 2021-12-09 at 11 57 27 AM

-----

## Supply liquidity

- [ ] Change the default view when you first open the supply liquidity tab:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 11 43 00 AM<br>(Don't know why it's showing as `invalid pair`?) | Screen Shot 2021-12-09 at 11 49 58 AM


- [ ] Update the general panel layout:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 12 01 38 PM | Screen Shot 2021-12-09 at 12 02 08 PM

- [ ] If a pool has Balanced rewards, we should display them:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 12 05 06 PM | Screen Shot 2021-12-09 at 12 06 00 PM

- [ ] If a pool doesn't exist yet, we should tell the user, and display the price ratio instead of the slider:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 12 07 28 PM | Screen Shot 2021-12-09 at 12 15 55 PM

For the message (not included in the code):
```
<p class="text-center label" style="padding-top: 100px;">There's no liquidity pool for this pair yet.<br>Supply liquidity for these assets in any quantity to create a new pool.</p>
```

- [ ] Update the confirmation modal for new liquidity pools:

**Current**  | **Design**
------------ | -----------
Screen Shot 2021-12-09 at 12 17 55 PM | Screen Shot 2021-12-09 at 12 26 43 PM

```
<div class="modal-content">
	<!-- Overview -->
	<h3 class="label text-center">Create liquidity pool?</h3>
	<p class="text-center">Send each asset to the contract,<br>then create the pool.</p>

	<div class="grid grid-top supply-asset">
		<div class="grid-cell text-center hide" style="margin-right: 35px;">
			<p style="margin-bottom: 15px;" class="label">Assets to send</p>
			<p class="text-bold value-supply-sicx" style="margin-bottom: 5px">10 sICX</p>
			<p><a href="#" class="button">Send</a></p>
		</div>
		<div class="grid-cell text-center" style="margin-right: 35px;">
			<p style="margin-bottom: 15px;" class="label">Assets to send</p>
			<img src="../../img/icon/tick-2.svg" style="margin-top: 12px;" width="25px">
		</div>
		<div class="grid-cell text-center">
			<p style="margin-bottom: 15px;" class="label">In contract</p>
			<p class="text-bold value-supply-sicx">10 sICX</p>
			<p><a href="#" class="text-red">Remove</a></p>
		</div>
	</div>

	<div class="grid grid-center supply-asset" style="margin-bottom: 30px;">
		<div class="grid-cell text-center" style="margin-right: 35px;">
			<p class="text-bold value-supply-bnusd" style="margin-bottom: 5px">0 bnUSD</p>
			<p><a href="#" class="button">Send</a></p>
		</div>
		<div class="grid-cell text-center">
			<p class="text-bold" style="margin-bottom: 5px">-</p>
		</div>
	</div>

	<!-- Fee -->
	<p class="label text-center hide">Your assets will be locked for 24 hours.</p>

	<!-- Modal actions -->
	<div class="modal-actions">
		<a href="#" class="cancel modal-close">Cancel</a> <a href="./" class="button disabled">Create pool</a>
	</div>
</div>
```

Don't have any custom tokens to test this with, so haven't been able to check the full user flow.

octocat512:
Refactoring checklist after this PR is finished:
- [ ] make a generalized VerticalDivider. 
- [ ] The Balanced Package is using the BigNumber type for handling big numbers. Instead of using BigNumber, what about using string type?
- [ ] Deprecate toLoop and toIcx function.
- [ ] Is there any solution to simply the LedgerConfirmMessage component?

avipatel91:
@octocat512 I've noticed that `CurrencyRow` and `CurrencySearch` both make calls to get balances. CurrencyRow for individual token (via `useCurrencyBalance`) and `CurrencySearch` for all tokens (via `useTokenComparator`). It seems that individual row  balances are already populated via polling. Is there a reason why a double request is being made? As in, why are all balances being requested again for sorting, when data is already provided?

Reason I ask this is because there is rearranging of rows when search modal first comes to view because the comparator is still getting the balances. To the user, they already see the balances (which were already populated through polling). This UX can confuse the user.

octocat512:
> @octocat512 I've noticed that `CurrencyRow` and `CurrencySearch` both make calls to get balances. CurrencyRow for individual token (via `useCurrencyBalance`) and `CurrencySearch` for all tokens (via `useTokenComparator`). It seems that individual row balances are already populated via polling. Is there a reason why a double request is being made? As in, why are all balances being requested again for sorting, when data is already provided?
> 
> Reason I ask this is because there is rearranging of rows when search modal first comes to view because the comparator is still getting the balances. To the user, they already see the balances (which were already populated through polling). This UX can confuse the user.

Thanks for your feedback. The issue also causes a lot of API calls. I am considering how I can solve the issue. Is there any good solution? @avipatel91 

Current Solutions: 
- memorizing rending `CurrencySearch` so the component won't be recreated when the token modal is toggled. In this case, the user can only see the flash the first time.
- aggregating tokens' balance function calls into one call. https://github.com/makerdao/multicall

lijac:
> The Balanced Package is using the BigNumber type for handling big numbers. Instead of using BigNumber, what about using string type?

The long-term goal is to remove BigNumber.
For now, many parts of the app are using BigNumber, so if we use string type, it would need additional string->BigNumber conversions.
Instead, we can use Currency/CurrencyAmount/Fraction to replace BigNumber.

avipatel91:
> > @octocat512 I've noticed that `CurrencyRow` and `CurrencySearch` both make calls to get balances. CurrencyRow for individual token (via `useCurrencyBalance`) and `CurrencySearch` for all tokens (via `useTokenComparator`). It seems that individual row balances are already populated via polling. Is there a reason why a double request is being made? As in, why are all balances being requested again for sorting, when data is already provided?
> > Reason I ask this is because there is rearranging of rows when search modal first comes to view because the comparator is still getting the balances. To the user, they already see the balances (which were already populated through polling). This UX can confuse the user.
> 
> Thanks for your feedback. The issue also causes a lot of API calls. I am considering how I can solve the issue. Is there any good solution? @avipatel91
> 
> Current Solutions:
> 
> * memorizing rending `CurrencySearch` so the component won't be recreated when the token modal is toggled. In this case, the user can only see the flash the first time.
> * aggregating tokens' balance function calls into one call. https://github.com/makerdao/multicall



> > @octocat512 I've noticed that `CurrencyRow` and `CurrencySearch` both make calls to get balances. CurrencyRow for individual token (via `useCurrencyBalance`) and `CurrencySearch` for all tokens (via `useTokenComparator`). It seems that individual row balances are already populated via polling. Is there a reason why a double request is being made? As in, why are all balances being requested again for sorting, when data is already provided?
> > Reason I ask this is because there is rearranging of rows when search modal first comes to view because the comparator is still getting the balances. To the user, they already see the balances (which were already populated through polling). This UX can confuse the user.
> 
> Thanks for your feedback. The issue also causes a lot of API calls. I am considering how I can solve the issue. Is there any good solution? @avipatel91
> 
> Current Solutions:
> 
> * memorizing rending `CurrencySearch` so the component won't be recreated when the token modal is toggled. In this case, the user can only see the flash the first time.
> * aggregating tokens' balance function calls into one call. https://github.com/makerdao/multicall

For the first solution, how will the balances in the modal be up-to-date? My assumptions could be wrong, but I am assuming you will still use underlying useTokenBalances hook via comparator? If that is the case, then how will that get triggered again? For instance, lets say the user toggles modal -> memoized -> toggle off -> execute trade -> toggle modal. 

Expected result: The balances should change in CurrencySearch
Actual result: previous balance is shown because of memoization.

You will also have to make changes to props, such as removing isOpen to prevent extra re-rendering and adding a balances prop (which would trigger re-render) if balances change.

For second solution, that will require contract support, right?

octocat512:
The ICX queue info shows wrong data. @lijac 
image

octocat512:
image

- [ ] ICX queue doesn't show the correct information.
- [ ] Pool share and Daily rewards are not calculated.

octocat512:
sejong custom token addresses: 
OMM: cxb94c0da9a8e650cc09b2f6b1bbfaa957ab47ec4c
GBET: cxc1996e56284899f372016a168f1bf1ffa8be547d
FIN: cxf9653740981c4f6f5aeba5646fd5ebc523ea488d

octocat512:
image
I created FIN/bnUSD liquidity by using the UI. The liquidity details section doesn't show the correct data for this new liquidity. @lijac

octocat512:
image

@parrot9design Adding a slider is tricky at this point. I'd like to implement the slider after PR is finished.

parrot9design:
> @parrot9design Adding a slider is tricky at this point. I'd like to implement the slider after PR is finished.

The slider needs to be added before we’ll approve this PR. The changes here are already a regression for people who just want to use an existing pool, let’s not make their experience even worse by removing the sliders as well.

lijac:
While testing, please be aware of the below glitch in sejong network.

On sejong network, sICX->ICX swap always (it doesn't matter if the user uses Swap field or For field) fails with the `InvalidParameter` error.
https://sejong.tracker.solidwallet.io/transaction/0x6be46527d405289dfa0975d9252cc4634b7fff9c2678c70040f8f05a1dde3f35

BALN->ICX also fails because it contains sICX->ICX route.
https://sejong.tracker.solidwallet.io/transaction/0x28c0f9306b97263b604c4ddcb3f176c473b83bfebc73d8a3051b18f45105e910

BennyOptions:
First thing I noticed, please change the modals to drop-down menus @lijac @octocat512 

On this PR
Screen Shot 2021-12-16 at 4 01 01 PM


How it should be
Screen Shot 2021-12-16 at 4 02 12 PM

BennyOptions:
I have been able to import tokens successfully and the supply liquidity flow seems good. However, I can't test the creation of a pool because I only have BALN, bnUSD and sICX. If somebody else can create a custom token on testnet and send me some, please let me know

parrot9design:
## Removing custom tokens

(Refer to the Swap > For section in the [development prototype](http://balanced-development.s3-website-us-east-1.amazonaws.com/prototype-dev/trade/) for the design.)

- [ ] Add the option to remove a custom token from the asset dropdown:
70FF0578-51B5-4DDB-B8BA-54C6DD63A808

The button should only appear on hover (except for on mobile devices, where it should always be visible). The rest of the time, it should just look like this:

ECD2645A-2761-4036-ACCE-34462929057F


- [ ] Add a confirmation modal:
0DEF600F-75C2-41D0-858E-BECF918206E2

```
<div class="modal modal-show" id="modal-import-asset">
			<div class="modal-content">
				<!-- Overview -->
				<h3 class="label text-center">Remove asset?</h3>
				<p class="text-bold text-center">GBET</p>
				<p class="text-center">Gangster Bet</p>

				<!-- Contract address -->
				<p class="label text-center" style="width: 175px;display: block;word-wrap: break-word;margin: auto;margin-top: 5px;">cxf61cd5a45dc9f91c15aa65831a30a90d59a09619</p>

				<!-- Modal actions -->
				<div class="modal-actions">
					<a href="#" class="cancel modal-close">Cancel</a> <a href="./" class="button red">Remove</a>
				</div>
			</div>
		</div>
```

BennyOptions:
Assuming we need to change the font size here

Screen Shot 2021-12-17 at 3 54 49 PM

avipatel91:
> 



> Assuming we need to change the font size here
> 
> Screen Shot 2021-12-17 at 3 54 49 PM

@BennyOptions What about "Search name or contract"

Pointing that out since that is what is shown in mocks.

octocat512:
please review and leave feedback. please add your comment to the below checklist.

- [ ] fix funding vote inputbox

parrot9design:
Fix the asset dropdowns. They should look the same as they do in the app now. Specifically:
- [ ] `Balance` -> `Wallet`
- [ ] Fix the width - it should be slightly wider than the asset field on each side

**Current** | **Design**
----------- | ------------
Screen Shot 2021-12-19 at 8 54 19 AM | Screen Shot 2021-12-19 at 8 52 10 AM

- [ ] Fix the length of the dropdown:

**Current** | **App**
----------- | ------------
Screen Shot 2021-12-19 at 8 54 19 AM | Screen Shot 2021-12-19 at 8 54 25 AM

We may need to limit it to 3 assets + the search field.

- [ ] Update the placeholder text to reflect the design. That's been addressed in #824 thanks to @avipatel91, but the font style needs to be updated to look like the design, and we need to re-add the search icon.

**Current** | **#824** | **Design**
----------- | ---------- | ------------
Screen Shot 2021-12-19 at 9 02 00 AM | Screen Shot 2021-12-19 at 9 02 24 AM | Screen Shot 2021-12-19 at 9 02 10 AM

- [ ] Update the supply liquidity UI to reflect the design

**Current** | **Design**
----------- | ------------
Screen Shot 2021-12-19 at 9 06 59 AM | Screen Shot 2021-12-19 at 9 09 04 AM

- [ ] Add the sliders back for exisiting pools

I can't test adding new tokens or liquidity pools, as the UI always returns "No results found" when I paste a contract address.

octocat512:
image
The price ratio is incorrect.

octocat512:
> I can't test adding new tokens or liquidity pools, as the UI always returns "No results found" when I paste a contract address.

Please check if the address contains space characters when pasting a contract address. It works fine for me. @parrot9design

parrot9design:
>Please check if the address contains space characters when pasting a contract address. It works fine for me. @parrot9design

Double-checked with several contract addresses, but still no luck. I pasted these addresses straight from the tracker, with no spaces:

BALW: `cxcfe9d1f83fa871e903008471cca786662437e58d`
CODA: `cxcf5424ca2acae98db83ee5a6a5fe8e8575474167`

----

Not sure why the scrollbar is appearing over the values here?
Screen Shot 2021-12-21 at 6 55 40 PM

----

The wallet values should be separated by commas every 3 digits for better readability (i.e. `4,002`, `400,002`, `4,000,002`)
Screen Shot 2021-12-21 at 6 55 58 PM

----

We should include the APYs for all BALN-incentivised pools where the rectangle is (see the [dev prototype](http://balanced-development.s3-website-us-east-1.amazonaws.com/prototype-dev/trade/)).
Screen Shot 2021-12-21 at 6 57 33 PM

----

I supplied liquidity to the BALN/bnUSD pool, but the data I see afterwards doesn't match up:
Screen Shot 2021-12-21 at 7 11 18 PM

- In the top section, it doesn't show a value under 'Your supply'
- 'Your daily rewards' in the top section is more than the total daily rewards, and the daily rewards shown in the liquidity details section

After going to a different page and coming back again, the data fixes itself, but it shouldn't require that action.
----

Can we add commas to the numbers shown in the liquidity details section?
Screen Shot 2021-12-21 at 7 11 18 PM

----

I still can't add a custom token, so I can't fully test that or creating a custom pool. Are there any tokens on Sejong we can test with?

octocat512:
sejong custom token addresses: (https://github.com/balancednetwork/balanced-network-interface/pull/807#issuecomment-995054414) @parrot9design 
OMM: cxb94c0da9a8e650cc09b2f6b1bbfaa957ab47ec4c
GBET: cxc1996e56284899f372016a168f1bf1ffa8be547d
FIN: cxf9653740981c4f6f5aeba5646fd5ebc523ea488d

octocat512:
> Double-checked with several contract addresses, but still no luck. I pasted these addresses straight from the tracker, with no spaces:
> 
> BALW: cxcfe9d1f83fa871e903008471cca786662437e58d
> CODA: cxcf5424ca2acae98db83ee5a6a5fe8e8575474167
Those are mainnet addresseses so those don't work on sejong testnet. @parrot9design

parrot9design:
- [ ] Fix the modal styles:

**Current** | **Design**
----------- | ------------
Screen Shot 2021-12-22 at 11 03 33 AM | 145302382-756c85f5-31ff-437b-ac8e-983d818ac6e8
Screen Shot 2021-12-22 at 11 32 56 AM | 

 
- The ticker should use `font-size: 20px`, `font-family: tex-gyre-adventor-bold`, `font-weight: normal`, `color: #ffffff`
- The asset name should use `font-size: 16px;`, `font-family: 'tex-gyre-adventor-regular';`, `font-weight: normal;`, `color: #ffffff;`
- The contract address should be evenly split over 2 lines
- The Import button should use `font-size: 14px`

------

- [ ] The remove button should only appear on hover (except for on mobile devices, where it should always be visible). The rest of the time, it should just look like this:

146511850-afdedd59-e030-4837-9b42-ecf07eb8a6df

----

- [ ] If you select a custom token and then remove it from the dropdown, it's still shown as the active asset after you confirm the removal:

Screen Shot 2021-12-22 at 11 42 30 AM


----

- [ ] Add `cursor: not-allowed` to the swap/supply buttons when you haven't entered an amount / don't have enough to complete the transaction: 

**Current** | **Design**
----------- | ------------
Screen Shot 2021-12-22 at 11 56 29 AM | Screen Shot 2021-12-22 at 12 03 11 PM

------

- [ ] The asset dropdowns don't look like they've been properly ordered yet:
Screen Shot 2021-12-22 at 12 21 41 PM

- They should be listed by highest-> lowest wallet balance, then alphabeticaly (for this example: sICX, ICX, bnUSD, BALN, OMM)
- When signed out, they should be listed alphabetically

-----

- [ ] If you expand an asset dropdown in a signed out state, then click the sign in button, the sign-in modal is behind the asset dropdown:

Screen Shot 2021-12-22 at 12 24 51 PM

-----

I have Sejong OMM in my wallet but it can only be detected on Omm, not Balanced or Hana, so I can't yet test supplying liquidity to a custom pool.

octocat512:
>  If you select a custom token and then remove it from the dropdown, it's still shown as the active asset after you confirm the removal:
Could you guide more details? @parrot9design

parrot9design:
>>If you select a custom token and then remove it from the dropdown, it's still shown as the active asset after you confirm the removal:

>Could you guide more details? @parrot9design

I made a video to help clarify:

https://user-images.githubusercontent.com/63842577/147163145-a774efd0-7d94-4e4a-9c78-a054cc8fe3e7.mov

As you can see at the end, after removing OMM from the dropdown, it's still shown as the active asset, and stays there even if I go to a different page and back again. It should probably default to an asset like BALN instead?

octocat512:
> > > If you select a custom token and then remove it from the dropdown, it's still shown as the active asset after you confirm the removal:
> 
> > Could you guide more details? @parrot9design
> 
> I made a video to help clarify:
> 
>  Screen.Recording.2021-12-23.at.11.38.14.AM.mov 
> As you can see at the end, after removing OMM from the dropdown, it's still shown as the active asset, and stays there even if I go to a different page and back again. It should probably default to an asset like BALN instead?

Setting token as default in the `CurrencyInputPanel` is difficult. For example, when a user tries to add OMM/BALN liquidity. After knowing The custom OMM address is not right, the user will remove it. In this case, if the frontend defaults the BALN token, then the frontend will set BALN/BALN liquidity. The front end can occur unexpected errors in this situation with the current logic. The below things are my suggestions.
1. set the default as an empty state. So the `CurrencyInputPanel` will show `Select a Token`.  
image
    I am not sure that this solution can affect other `CurrencyInputPanel` in swap and voting proposal sections. 
2. We can skip the issue. We can find a good solution later. I think it isn't important for now.
My preference is second. 😂 @parrot9design

parrot9design:
> 2\. We can skip the issue. We can find a good solution later. I think it isn't important for now.
>         My preference is second. 😂 @parrot9design

Let's skip it, I'm not super fond of showing text in the input panel. 😅

parrot9design:
Now that it's on mainnet, I tested with BALW and CODA:
Screen Shot 2021-12-23 at 4 18 25 PM Screen Shot 2021-12-23 at 4 19 09 PM

We should limit the width the asset name can take up so it wraps over 2 lines instead of getting hidden behind the button, and prevent the button label from being spread across 2 lines.

parrot9design:
This is a new pool, so the button label should say `Create pool` instead of `Supply`.
Screen Shot 2021-12-23 at 4 25 19 PM

----

I thought the 4 assets you can create a pool against are sICX, bnUSD, IUSDC, and USDS, not BALN. @BennyOptions is that right?
Screen Shot 2021-12-23 at 4 27 19 PM

BennyOptions:
Found an issue with withdrawing liquidity


https://user-images.githubusercontent.com/43685868/147185806-89df2d41-ec27-4ef9-be2c-9a62e2b6c328.mov

BennyOptions:
> I thought the 4 assets you can create a pool against are sICX, bnUSD, IUSDC, and USDS, not BALN. @BennyOptions is that right?

Yes, that's correct. @octocat512 please remove BALN as an option for the bottom drop-down menu and add IUSDC

BennyOptions:
I successfully created a pool with no issue

BennyOptions:
Supply liquidity flow is very buggy right now


https://user-images.githubusercontent.com/43685868/147188495-b9602395-79f3-4017-8214-bb6cbc7f614b.mp4

parrot9design:
Just found a couple more issues:

Screen Shot 2021-12-24 at 11 41 34 AM

- After creating a liquidity pool, the top panel doesn't indicate that it's now a pool.
- If the pool share is 100%, we don't need the decimal places as it can't get any higher.

---

Screen Shot 2021-12-24 at 11 40 30 AM

- The banner messages only indicate supplying liquidity. Let's use these ones for a new pool:

> Creating Asset1 / Asset2 pool...

> Asset1 / Asset2 pool created. x Asset1 and y Asset2 supplied.

> Couldn't create Asset1 / Asset2 pool.

---

I haven't put the flow through its paces enough to see if Scott's issues are resolved, but from a UI/UX perspective, this is looking really nice now. Awesome to see the APY in place for the BALN pools, too. Thanks so much for all your hard work, @octocat512 and @lijac. 🙏👏

avipatel91:
A loading spinner is missing for the liquidity pool panel. For slow network connections, users will see an empty panel on initial render.

https://user-images.githubusercontent.com/11547815/147301293-ae2370d2-3aa6-464f-ac17-918a3210802b.mov

parrot9design:
We'll need to check the messages in the activity history section, as well. Just spotted this:
Screen Shot 2021-12-24 at 1 30 34 PM

octocat512:
Refactoring list:

- [ ] Remove `pool/hooks.ts` file and `pool/hooks.ts.old` and Use `useV2Pairs` hook instead of  `uesPools` hook. Remove `PoolState` interface.

BennyOptions:
Small issue, but please line up the arrows

Screen Shot 2021-12-27 at 8 29 01 PM

BennyOptions:
Supply liquidity flow still seems slow and glitchy. Any ideas on how to improve it? If this is too much to fix, we can deploy now and generally work on code refactor to increase performance. Performance of the app seems slow


https://user-images.githubusercontent.com/43685868/147518076-63d4817c-b08e-4268-a2f7-ad27ab08557b.mov

BennyOptions:
Another example of performance issues


https://user-images.githubusercontent.com/43685868/147518187-1f9d3628-cd47-4695-b315-bf5a2f2bf286.mov

BennyOptions:
APIs also slower to load when changing from Home to Trade

Screen Shot 2021-12-27 at 8 37 39 PM

BennyOptions:
Another example of slow API responses. This does not happen on the production build

Screen Shot 2021-12-27 at 8 39 20 PM

BennyOptions:
Also, when using a Ledger, it often takes much longer for the transaction to appear on the Ledger device as compared to the production build. On production, when I click "supply" on the webapp, I immediately look down at my ledger and it is asking "approve" or "reject". 

On this build, it is often a few seconds before it appears on my Ledger device to confirm

lijac:
Close #830

BennyOptions:
This is much better @octocat512, great work!

PR #810, opened 2021-11-30
feat: reposition notification for small devices

hetfly:


parrot9design (approved):
Much better, thanks @hetfly!

octocat512:
Instead of using `useMedia` hook, what about using `react-device-detect`? @hetfly 

https://github.com/balancednetwork/balanced-network-interface/blob/4381d64abb7886c7887506080f4686cd84dbf3f0/src/app/components/Modal/index.tsx#L5

hetfly:
@octocat512  I think this is more of a screen size condition. Under 1000px the notification overlaps with the app navigation. That can happen on resized laptop app window as well.

octocat512:
Make sense @hetfly

PR #812, opened 2021-12-01
feat: add proper empty states

hetfly:


hetfly:
@parrot9design empty states updated but I was unable to simulate the wrong sizing, do you still experience this issue ?

parrot9design:
>I was unable to simulate the wrong sizing, do you still experience this issue ?

It looks like it's been fixed, I'm no longer able to replicate it. 👍

parrot9design (approved):

PR #823, opened 2021-12-15
build: add commitizen to help create conventional commits

### LINKED ISSUE #460: Use these development methods
git commit message convention
https://www.conventionalcommits.org/en/v1.0.0/#specification

version control or distribution convention
https://semver.org/

avipatel91:
Commitizen has been added and hooked into git via husky.

We use [cz-conventional-changelog](https://www.npmjs.com/package/cz-conventional-changelog) adapter. 
We can modify the configuration through package.json to change settings such as max lengths, capitalization, etc...

To use Commitizen, first install dependencies. Then simply type `git commit`. This will prompt the dev
for info in order to generate a commit message.

Another option is [this](https://github.com/streamich/git-cz), which allows emojis.
Resolves #460

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn run checkTs && lint-staged"
+      "pre-commit": "yarn run checkTs && lint-staged",
+      "prepare-commit-msg": "exec < /dev/tty && git cz --hook || true"
     }
   },
   "jest": {
@@ -163,5 +164,13 @@
       }
     }
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "commitizen": "4.2.4",
+    "cz-conventional-changelog": "3.3.0"
+  },
+  "config": {
+    "commitizen": {
+      "path": "./node_modules/cz-conventional-changelog"
+    }
+  }
 }


shallcdn:
I have tried with these npm packages; I think that's really good for new members or some members who don't have familiarity with the commit message convention.

BennyOptions:
@octocat512 should this be merged? I just noticed the comment by @shallcdn and it sounds good, but I don't fully understand the impact.

robcxyz:
So I am not very familiar with commitizen but from what I can tell, it is mainly focused on enforcing the commit messages themselves and then through the adapters updating things like the changelog within the commit itself correct?  Also not sure how it handles release drafts but I wanted to propose an alternative solution based on GH actions that does both the changelog and release draft. I think commitizen can be used with it for those that don't know 

Been evaluating a few GH actions out there and I personally think [google-github-actions/release-please-action](https://github.com/google-github-actions/release-please-action) is the best one IMO out of the others I have checked out ([1](https://github.com/TriPSs/conventional-changelog-action), [2](https://github.com/bcoe/conventional-release-labels). It creates the changelog based on `feat` / `fix` commit prefixes but unlike other GH actions I have evaluated, it does not force a semvar version bump based on each commit and instead groups them into a PR so that when you are ready to do a release, you can have multiple changes bundled into one release. It also automates the bumping of versions so that if you have the version hardcoded into a package.json, it will bump that for you on release. Handles all the languages in use in our ecosystem (ie js, python, java) for bumping various types of version files. 

I think the ideal solution would be to use this GH action and recommend to users who are new to conventional commits to use commitizen. Most merges into this repo are based on PRs as well so if a user doesn't format their commit messages correctly, the commits can be squashed with a message that conforms to the conventional commit standard.

avipatel91:
Like you mentioned, Commitizen just handles message format. It solves one of the two items in the issue. For changelog and versioning, GH action you linked is ideal.

robcxyz:
@avipatel91 - Yeah agree. I think BTP might be adopting this into their release pattern but still evaluating to see if it can be used within monorepo release patterns. But I think it would look good if all the changelogs / releases looked the same.

BennyOptions:
@hetfly @shallcdn I am cleaning up the repo and wondering if we can close this PR. Is there any next step? Blocker? Let me know if we can merge or close.

PR #831, opened 2021-12-30
fix: add missing minimumReceive to swapICX

### LINKED ISSUE #830: Minimum to receive not working properly
The frontend did not properly pass the Minimum to Receive number to the blockchain.

Transaction:
https://tracker.icon.foundation/transaction/0x5a6b07b3c5fa3613627f5e65ae10aaccdff2f8aa72e5e8c8784ebaa6475c17a0

lijac:
Close #830


lijac:
Now swaps are of 2 kinds:
1. starting from ICX, 
    It calls `router::route` function directly
2. starting from tokens
    It sends tokens to `router` contract and `router`’s `tokenFallback` is called

There was a bug implementing 1, it didn't pass `minReceive` param to `route` method thus setting it 0 by default.
As a fix, the `router` contract call (in the frontend) has been updated to include `minReceive` param.

PR #840, opened 2022-01-08
add: optimise small device ux

hetfly:


parrot9design:
- [ ] Love the way you've handled the wallet. For the BALN entry, I wonder if we remove the $ value from `Available` at mobile sizes so it looks like this?
Screen Shot 2022-01-18 at 10 02 28 AM

---
Collateral/Loan section:

- [ ] Can we keep the Confirm/Cancel buttons side by side for the Collateral section until at least 360px? 
- [ ] When the screen is reduced to 280px, the Cancel/Confirm buttons for the Loan section are pushed to the centre. Even though we likely won't have many people crazy enough to use Balanced on a device this small, we should keep the buttons right-aligned.
- [ ] When the Confirm/Cancel buttons are displayed one on top of the other, the Collateral section positions Confirm above Cancel, while Loan does the opposite. We should keep it consistent. I'm thinking Cancel should go above Confirm, because in a situation like this, it's harder to tell that Cancel is actually a button:
Screen Shot 2022-01-18 at 10 21 41 AM

---

- [ ] For the Activity section, now that we have horizontal scrolling, can we add the Amount column? Certain actions, like staking BALN, don't show the amount in the activity message, so this will allow people to get the full context.

---

A few other things I missed in the initial issue:

- [ ] Can we reposition the external link icons, so they're sitting on the same line as the text?

**Current** | **[Dev prototype](http://balanced-development.s3-website-us-east-1.amazonaws.com/prototype-dev/)**
----------- | ---------------
Screen Shot 2022-01-18 at 9 51 45 AM | Screen Shot 2022-01-18 at 9 52 32 AM


- [ ] When a vote is live, the Approve/Reject buttons are huge on mobile. Let's move them side by side underneath, similar to how it looks _after_ you vote:
 

- [ ] Fix the way the page header wraps when the proposal type is too long. The specific proposal type should wrap over 2 lines, not the header:
Screen Shot 2022-01-18 at 10 05 24 AM

- [ ] Fix the alignment for the labels/input values for `BALN allocation` and `Network fee allocation`:
Screen Shot 2022-01-18 at 10 08 51 AM
- Centre the labels with the input boxes
- Add some padding between the label and input box
- If possible, align the bottom of the input boxes with the bottom of the current value on the left

- [ ] For the Funding proposal, add comma formatting to the values (sorry, I know this isn't technically a responsive issue 😅)
Screen Shot 2022-01-18 at 10 06 07 AM

hetfly:
@parrot9design updated, except of the case with width < 280px as we discussed it's not worth handling.

parrot9design (approved):
This is so much better, thanks @hetfly! 🙏 The only thing I can't test right now are the vote buttons, but we can address that later if it's still an issue for the next proposal.

src/app/components/Form/index.tsx
```diff
@@ -27,14 +38,18 @@ const CurrencyInput = styled(Box)`
...
+  padding: 3px 7px;
   height: 40px;
   border-radius: 10px;
   outline: none;
   width: 100%;
   -webkit-appearance: none;
   transition: border 0.3s ease;
   overflow: visible;
+
+  @media screen and (min-width: ${smallScreenSize}) {
```
octocat512, line 50: The current breakpoints in the theme file don't fit the different devices? @hetfly 
If it is not fit, then I'd like you to define the breakpoints and use them.

octocat512 (changes requested):

src/app/components/Form/index.tsx
hetfly: updated, thanks!

hetfly:
@parrot9design you can review voting buttons now

parrot9design:
> @parrot9design you can review voting buttons now

Looks great, thanks @hetfly!

octocat512:
Please resolve the conflicts @hetfly

src/app/components/Pagination/index.tsx
```diff
@@ -28,6 +28,10 @@ const NumberButton = styled(Button)`
   &.active {
     background: ${({ theme }) => theme.colors.primary};
   }
+
+  @media screen and (max-width: 340px) {
```
octocat512, line 32: please use defined breakpoints

src/app/components/Tooltip/index.tsx
```diff
@@ -17,8 +17,8 @@ export const TooltipContainer = styled.div<{ wide?: boolean; small?: boolean; wi
     ${props => props.small && ' width: 180px; padding: 11px;'}
   }
 
-  @media (max-width: 410px) {
-    ${props => props.small && ' width: 120px; padding: 10px'}
+  @media (max-width: 370px) {
```
octocat512, line 20: please use defined breakpoints

octocat512 (changes requested):

Issue #833, opened 2022-01-03
Network Fee Claim Logic


BennyOptions:
The smart contract can handle a trailing 50 days of claims. Let's set the Claim button to send a transaction to claim a range from currentDay to currentDay - 49 (let's do 49 instead of 50 to be safe)

The current UI seems to not bundle in 50 days

BennyOptions:
Issue

Current Logic: Frontend fetches
1 - 50, 51-100, 101-150, 150-200, 201-250, 251-254


goal:
254-204, 204-154....

PR #847, opened 2022-01-11
Fix: update network fee logic

Chunn:


octocat512 (changes requested):
While developing, You don't use Estlint? Please lint before committing.

src/app/components/home/RewardsPanel.tsx
```diff
@@ -205,15 +205,17 @@ const NetworkFeeSection = () => {
 
   const changeShouldLedgerSign = useChangeShouldLedgerSign();
   const addTransaction = useTransactionAdder();
+
+  const DAY_LIMIT = 49;
```
octocat512, line 209: Could you use the constant variable `BATCH_SIZE`? @Chunn 
Did you confirm that the UI claims the same amount same as the showing amount?

octocat512 (changes requested):

octocat512:
> > While developing, You don't use Estlint? Please lint before committing.
> 
> @octocat512 Sorry about that 🙇‍♂️. I work on Win. It seems there are some problem with estlint on Window 🙈

I am using Windows too. But no problem with eslint on Windows. If you have issues, I can help you. @Chunn

Chunn:
> > > While developing, You don't use Estlint? Please lint before committing.
> > 
> > 
> > @octocat512 Sorry about that 🙇‍♂️. I work on Win. It seems there are some problem with estlint on Window 🙈
> 
> I am using Windows too. But no problem with eslint on Windows. If you have issues, I can help you. @Chunn

Oh. Thank you! It seems I've solved it

src/app/components/home/RewardsPanel.tsx
Chunn: @octocat512 I've used `DAY_LIMIT` instead of `BATCH_SIZE` based on this comment https://github.com/balancednetwork/balanced-network-interface/issues/833#issue-1092830176
> Did you confirm that the UI claims the same amount same as the showing amount?

 I don't have an account on mainnet. So, I've asked @rise-techiast  to check for me. It seems it's OK

BennyOptions:
@rise-techiast please follow up with @Chunn

src/app/components/home/RewardsPanel.tsx
rise-techiast: We've discussed and agreed to use `BATCH_SIZE` only

BennyOptions:
@Chunn @rise-techiast please disregard my comment about using 49 from the issue. Please use the BATCH_SIZE variable. Do not use DAY_LIMIT.

octocat512:
https://github.com/balancednetwork/balanced-contracts/blob/development/core_contracts/dividends/dividends.py
Please reference the smart contract code if you can read python. All API calls are failing with the error message. `Start must not be greater than or equal to end.` @Chunn

Chunn:
> https://github.com/balancednetwork/balanced-contracts/blob/development/core_contracts/dividends/dividends.py Please reference the smart contract code if you can read python. All API calls are failing with the error message. `Start must not be greater than or equal to end.` @Chunn
@octocat512 As the task description, we'll have start >= end. It seems smart contract code and FE code don't have logic consistent.
Please refer: https://github.com/balancednetwork/balanced-network-interface/issues/833#issue-1092830176
@BennyOptions Could you confirm the logic for this? Thank you!

BennyOptions:
@Chunn I think there is some confusion

On the smart contract, the maximum batch size to get data is 50 entries (i.e. entry 1-50, 51-100, etc.). 

Smart contract method:
getUserDividends: [user address, start, end]

Start must be lower than end (i.e start = 10, end = 12), and start can **never be** more than 50 away from end. That is the constraint.

Current FE Logic:
Currently, the FE does checks starting at 1-50. First is getUserDividends: [user address, 1, 50], then second is getUserDividends: [user address, 51, 100], etc. etc. until we reach the current day number.

**Correct FE Logic:**
The correct FE logic should start from the current day number, and work backwards. According to stats.balanced.network, current day right now is 317. So, the FE first should should be getUserDividends: [user address, 267, 317], then next should be getUserDividends: [user address, 216, 266]. It should repeat this process until start = 1, removing 50 days each time.

In summary, the logic is **not** start >= end, the logic always is end >= start as consistent with the smart contract. The issue is about how the FE creates batches of 50.

If you still have questions, please let me know, thanks!

Chunn:
> @Chunn I think there is some confusion
> 
> On the smart contract, the maximum batch size to get data is 50 entries (i.e. entry 1-50, 51-100, etc.).
> 
> Smart contract method: getUserDividends: [user address, start, end]
> 
> Start must be lower than end (i.e start = 10, end = 12), and start can **never be** more than 50 away from end. For example, getUserDividends: [user address, 1, 100] and getUserDividends: [user address, 10, 5] will both break. That is the constraint.
> 
> Current FE Logic: Currently, the FE does checks starting at 1-50. First is getUserDividends: [user address, 1, 50], then second is getUserDividends: [user address, 51, 100], etc. etc. until we reach the current day number.
> 
> **Correct FE Logic:** The correct FE logic should start from the current day number, and work backwards. According to stats.balanced.network, current day right now is 317. So, the FE first should should be getUserDividends: [user address, 267, 317], then next should be getUserDividends: [user address, 216, 266]. It should repeat this process until start = 1, removing 50 days each time.
> 
> In summary, the logic is **not** start >= end, the logic always is end >= start as consistent with the smart contract. The issue is about how the FE creates batches of 50.
> 
> If you still have questions, please let me know, thanks!

@BennyOptions I got it. Thanks for clarifying that 🙇

shallcdn (commented):
@Chunn Please check my comments!

shallcdn (approved):
LGTM!

PR #852, opened 2022-01-14
Hotfix/remove bn usd base token selection

### LINKED ISSUE #849: Base/Quote issues in UI
For the bottom token, only the following tokens should be allowed:

- bnUSD
- sICX
- USDS
- IUSDC

As you can see, CFT is here

Also, please don't allow bnUSD to be at the top. Right now, it is causing lots of errors to the community. There is no logical reason to have bnUSD at the top anyway.

octocat512:
close  #849


BennyOptions (approved):
fixes the issue

PR #862, opened 2022-01-24
fix: update liquidation price tooltip

### LINKED ISSUE #861: Update the liquidation tooltip
Reverse rebalancing will cause the liquidation price to increase, which can get people in a dangerous situation because they think they have more leeway during a market correction than they actually do. It also wasn't clear that your _entire_ collateral position would be liquidated. This message should fix that.

Screen Shot 2022-01-24 at 9 58 21 AM

```
<div class="tooltip-content" style="min-width: 270px;">
	<p>If the ICX price reaches <span class="value-liquidated text-bold">$1.3126</span>, all your collateral will be liquidated.<br><span class="label" style="font-size: 12px;">Keep a close eye on the price, as rebalancing may cause it to fluctuate.</span></p>
</div>
```

octocat512:
close #861

### DIFF
--- a/src/app/components/home/PositionDetailPanel.tsx
+++ b/src/app/components/home/PositionDetailPanel.tsx
@@ -268,8 +268,15 @@ const PositionDetailPanel = () => {
           </Box>
 
           <Tooltip
-            text={`If the ICX price reaches $${liquidationThresholdPrice.toFixed(3)}, 
-                    your collateral will be liquidated.`}
+            text={
+              <Typography variant="body">
+                If the ICX price reaches ${liquidationThresholdPrice.toFixed(3)}, your collateral will be liquidated.{' '}
+                <br />
+                <Typography as="small" fontSize={12} color="text1">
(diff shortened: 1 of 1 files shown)


src/app/components/home/PositionDetailPanel.tsx
```diff
@@ -268,8 +268,15 @@ const PositionDetailPanel = () => {
           </Box>
 
           <Tooltip
-            text={`If the ICX price reaches $${liquidationThresholdPrice.toFixed(3)}, 
-                    your collateral will be liquidated.`}
+            text={
+              <Typography variant="body">
+                If the ICX price reaches ${liquidationThresholdPrice.toFixed(3)}, your collateral will be liquidated.{' '}
```
parrot9design, line 273:
```suggestion
                If the ICX price reaches ${liquidationThresholdPrice.toFixed(3)}, all your collateral will be liquidated.{' '}
```

PR #867, opened 2022-01-25
Feat/update risk ratio bar

### LINKED ISSUE #865: Update the risk ratio bar
We shouldn't imply that a position of any size is "low risk". 

Let's remove that label and its tooltip, and merge it with the larger bar. You'll need to take care to optimise for smaller positions, as the logic of the low risk section is slightly differt to make sure a value is displayed on the bar for all position sizes.

Screen Shot 2022-01-25 at 5 02 09 PM

hetfly:
close #865


parrot9design (approved):
This works well, thanks @hetfly! I noticed that I couldn't max out the borrow slider, but that may be due to the change in LTV?

Screen Shot 2022-01-26 at 1 32 46 PM

PR #882, opened 2022-02-08
Revert "Revert "add: optimise small device ux""

octocat512:
Reverts balancednetwork/balanced-network-interface#881


octocat512:
please continue to work on the PR. @hetfly

parrot9design:
The mobile experience is so smooth now, nicely done. Just a few final things I noticed:

- [ ] The risk ratio tooltips overlap in several positions as you scroll down the page:
  

- [ ] Let's move the liquidity APY underneath the pool (and hide the `:`) to avoid it wrapping like this:

parrot9design:
This is good! Let's just tidy up the liquidation tooltip (content and width) and this should be good to ship.

**Current** | **Ideal**
----------- | -----------
 | Screen Shot 2022-01-27 at 3 27 33 PM



>If the ICX price reaches **$x.xxx**, all your collateral will be liquidated.
>Keep a close eye on this number, as rebalancing may cause it to fluctuate.

parrot9design (approved):

Issue #884, opened 2022-02-09
Liquidity Details: Hide small values


BennyOptions:
If a user owns 100% of a pool, it is impossible to remove all liquidity because of the smart contract rules. However, we can hide small values in the UI to prevent clutter for the user. This only effects users that own 100% of a pool.

An example is below. GBET/USDS pool was created as a test. Now, it will always be there because removing 100% is impossible. It would be nice to have it disappear.

![Screen Shot 2022-02-09 at 2.17.28 PM.png](https://images.zenhubusercontent.com/5f07ce5a8cf129080163e5a4/430d11cf-1638-445d-b44b-6f5a3b0771cb)

PR #895, opened 2022-02-19
Feat: hide pool record if base balance of a pool is too small

shallcdn:
- [x] Add logic to check if the  pool share value less than 0.01 percent, the pool record will be hidden

Ref: #884


src/app/components/trade/LiquidityDetails.tsx
```diff
@@ -219,19 +220,25 @@ const PoolRecord = ({
...
-        {`${bBalance.toFixed(2, { groupSeparator: ',' }) || '...'} ${bBalance?.currency.symbol || '...'}`}
-      </DataText>
-      {upSmall && <DataText>{`${share.multiply(100).toFixed(4) || '---'}%`}</DataText>}
-      {upSmall && <DataText>{`~ ${reward.toFixed(4, { groupSeparator: ',' }) || '---'} BALN`}</DataText>}
-      <DataText>
-        <WithdrawText pair={pair} balance={balance} poolId={poolId} />
-      </DataText>
-    </ListItem>
+    <>
+      {parseFloat(share.multiply(100).toFixed(4)) > LOWEST_LIQUIDITY_POOL_PERCENT ? (
```
lijac, line 224: Can we use comparison methods of `Fraction` directly? @shallcdn

lijac: `share.multiply(100).greaterThan(LOWEST_LIQUIDITY_POOL_PERCENT_FRAC)`
and
`LOWEST_LIQUIDITY_POOL_PERCENT_FRAC = new Fraction(1, 100)`

shallcdn: @lijac  Yes, we can, but I wonder why don't we use pure JS as the way I wrote the codes? Do we have any reasons? 
Ext: easy to read, consistency... I just want to know the reason to prevent my mistakes for any PRs in the future.

lijac: Sure, here are the reasons I can think of:
1. We may need to update `Fraction` class in the future and that can be `toFixed` method. If that's the case, we will probably update `greaterThan` to accommodate new changes but cannot guarantee `toFixed(4)` still returns the desired value.
2. We also have `BigNumber` in the project for processing numbers (though it is going to fade out), it's better to use explicit functions of separate classes to prevent confusion.

Personal preference also has a place in my mind :)

shallcdn: @lijac I have updated, Please help take a look again, thank you!

lijac:
@BennyOptions I assume we don't hide ICX/sICX pool even if pool share % is less than 0.01%. Can you confirm it?

BennyOptions:
@shallcdn my apologies, I should have shared more detailed logic. Using the pool-share percentage is dangerous because average users participating in a large pool will likely have small percentages of the total pool and their contribution will be hidden.

Please use a different logic to hide small values. I would prefer this logic:

If (quoteDepositAmount < 0.01 && baseDepositAmount < 0.01)
  then hidePool

octocat512:
Instead of hiding the small amount, I'd like to display the portions as lesser than 0.01%.
for example: <0.01%

BennyOptions:
https://github.com/balancednetwork/balanced-network-interface/pull/895#issuecomment-1054596981

We will use the above linked solution for now, and base it on **right-side currency (quote currency)**. So if quote currency amount is less than 0.01, hide it.

shallcdn:
> [#895 (comment)](https://github.com/balancednetwork/balanced-network-interface/pull/895#issuecomment-1054596981)
> 
> We will use the above linked solution for now, and base it on **right-side currency (quote currency)**. So if quote currency amount is less than 0.01, hide it.

Hi @BennyOptions let me confirm again your comment:
Screen Shot 2022-03-02 at 17 35 24

With the logic 
```
If (quoteDepositAmount < 0.01 && baseDepositAmount < 0.01)
then hidePool
```
I understand your mean and based on the image above:

_quoteDepositAmount_ stand for sICX value
_baseDepositAmount_ stand for bnUSD value

Isn't it?

>  and base it on right-side currency (quote currency). So if quote currency amount is less than 0.01, hide it.

so which column's value do you want to mention?

BennyOptions:
@shallcdn my apologies, I always get base and quote confused. I will speak only in terms of left-side currency and right-side currency. As an example for sICX / bnUSD, sICX = left-side currency and bnUSD = right-side currency.

The logic I would like should be based only on right-side currency. The reason is that there are constraints at the smart contract level around which currencies are allowed to be right-side currencies. This means we can be comfortable in choosing nominal constraints for right-side currency, such as < 0.01. The specific logic I recommend is below, with the understanding that baseDepositAmount represents the right-side currency:

```
If (baseDepositAmount < 0.01)
then hidePool
```

shallcdn:
@lijac li I have updated the logic of this MR based on the review of @BennyOptions. Please help retake a look at one.

one more thing is that the function greaterThan doesn't work as we expected, so I changed to use it as usual way comparison

shallcdn:
@lijac l have updated logic based on the [comment](https://iconcommunity.slack.com/archives/G0117R69JK0/p1646424448973689?thread_ts=1645848889.366999&cid=G0117R69JK0 ) and squash commit
Please help take a look. Thank you!

lijac (approved):
LGTM, ready to merge.

BennyOptions:
ready to merge - i checked UX

eulerbeat:
Close #909

PR #908, opened 2022-02-28
feat: add keyboard control to currency select

hetfly:


hetfly:
@parrot9design focusing input on dropdown show isn't happening on mobile since the keyboard would automatically popup

parrot9design:
This looks good, thanks @hetfly!

The only issue I noticed is that the search doesn't always work as expected. Examples:
Screen Shot 2022-03-08 at 5 39 27 PMScreen Shot 2022-03-08 at 5 39 35 PM
Screen Shot 2022-03-08 at 5 40 00 PMScreen Shot 2022-03-08 at 5 39 42 PM

It's not a result of this specific PR (and we can tackle it separately if that makes more sense), but are you able to take a look at this? Both the full token name and the ticker should be searchable.

hetfly:
@parrot9design opened new PR https://github.com/balancednetwork/balanced-network-interface/pull/916

parrot9design (approved):

src/app/components/SearchModal/CurrencyList.tsx
```diff
@@ -33,6 +36,9 @@ function CurrencyRow({
   showCurrencyAmount?: boolean;
   onRemove: () => void;
   account?: string | null;
+  isFocused: boolean;
+  index: number;
+  setFocusedIndex: React.Dispatch<React.SetStateAction<number | undefined>>;
```
octocat512, line 41: use `onFocus` instead of `setFocusedIndex`, remove index, and define the function like it.
`onFocus: () => void`

src/app/components/SearchModal/CurrencyList.tsx
```diff
@@ -111,6 +130,9 @@ export default function CurrencyList({
             setRemoveToken(currency as Token);
             showRemoveView();
           }}
+          isFocused={index === focusIndex}
```
octocat512, line 133: Instead of handling index in child components, handle here.
`onFocus = { () => setFocusIndex(index) }`

src/app/components/SearchModal/CurrencySearch.tsx
```diff
@@ -143,11 +141,55 @@ export function CurrencySearch({
...
 
+  const currencies =
+    currencySelectionType === CurrencySelectionType.NORMAL ||
+    currencySelectionType === CurrencySelectionType.TRADE_MINT_BASE
+      ? filteredSortedTokensWithICX
+      : filteredSortedTokens;
+
+  // keyboard control
+  const enter = useKeyPress('Enter');
+  const escape = useKeyPress('Escape');
```
octocat512, line 170: Move keyboard control logic in the `CurrencyList` component.

octocat512 (changes requested):

hetfly:
refactored, thanks for the review! @octocat512

octocat512 (approved):

PR #916, opened 2022-03-08
Feat/optimise token search

hetfly:


hetfly:
@parrot9design 

1) The inconsistencies mentioned in https://github.com/balancednetwork/balanced-network-interface/pull/908 were caused by searching query in the beginning or in the end of the token name/symbol. Changed it to searching anywhere in name/symbol. Have a look and let me know if that's better.

2) Also added new field to ```Token``` for searching helpful terms. This can fine tune the search for cases like 'dollar', 'optimus'. Also might be useful for grouping like 'stable', or in the future for cross chain reference.

parrot9design (approved):
Nicely done, @hetfly!

octocat512 (approved):

Issue #898, opened 2022-02-22
Ledger confirmation appears in two spots


BennyOptions:
See below screenshot. It should not appear in the "rewards" tab. It should only appear on the modal.
<img width="1059" alt="Screen Shot 2022-02-22 at 12 29 21 AM" src="https://user-images.githubusercontent.com/43685868/155092854-c2794e2b-80d8-46d5-8ca6-138ca90d7d24.png">

PR #917, opened 2022-03-08
fix: remove redundant ledger confirm message

shallcdn:
- [x] Remove leger confirm message on RewardsPanel

Ref #898


lijac (commented):
LGTM

PR #922, opened 2022-03-16
Feat/upgrade trading view library

### LINKED ISSUE #889: Upgrade the Trading View Library

hetfly:
close #889


BennyOptions:
@hetfly can u give me an approximate timeline for completing this task? Just trying to think about priorities and next steps

hetfly:
@BennyOptions should be done early next week

BennyOptions:
blocked by backend requirement to save charts made by users

parrot9design:
Is it possible to apply the TV timeframe to the Balanced chart when you close the modal, like you've done with the swap pair?

---

If you search for a chart without adding a `/` between the symbols, no results appear:

Screen Shot 2022-04-21 at 4 50 32 PM

---

On mobile, we might need to reduce the modal height to make it easier to close out. If the browser chrome has been hidden (standard behaviour when scrolling a page), this is how the TV chart appears when it first opens:


But if the chrome isn't hidden, or you tap the chart or the space up the top to close the modal, it looks like this:
 


It makes it much harder to close out of the chart without just refreshing the page, and difficult to change pairs. Tapping the symbol search also triggered the browser chrome which often moved search (or the ability to close it) out of reach.

parrot9design (approved):

parrot9design:
This PR is no longer blocked, as we’ll ship the outstanding functionality in #961.

@octocat512 let’s get this reviewed and out to the community. 💪

lijac:
Wonderful! Absolutely stunning!

A few comments:

- Can we customize `undefined` in `USDS/bnUSD, undefined, 1h`? That seems to be the name of the data provider (service).

image

parrot9design:
> @parrot9design what are your thoughts on releasing without `5m` and `15m` resolutions in TV ?

Happy to omit the 5m and 15m views. It sounds like more trouble than it’s worth to support it, and there’s limited value zooming in below 1h anyway.

lijac:
Looks good. The resolution sync TradingView -> Swap doesn't work on my side.
Except that, I think it's good to merge.

shallcdn:
Great to see it. Nice to have more information in this popup
Screen Shot 2022-04-30 at 00 12 34

The rest LGTM!

hetfly:
@lijac 
> The resolution sync TradingView -> Swap doesn't work on my side. Except that, I think it's good to merge.
That's right, it shouldn't work.

> This will work the same on test/production since the `API_ENDPOINT` is the same for both environments, right? @hetfly
Correct

@shallcdn 
> Nice to have more information in this popup
There really is not much to add except of Balanced exchange that would be repeated for every symbol at the moment. This is more suited for traditional exchange assets.

lijac (approved):

PR #935, opened 2022-03-31
String updates

parrot9design:


src/store/transactions/hooks.tsx
```diff
@@ -49,7 +49,7 @@ export function useTransactionAdder(): (
         onClick: () => window.open(link, '_blank'),
       };
 
-      toast(<NotificationPending summary={pending || t`Your transaction has been sent to the network`} />, {
+      toast(<NotificationPending summary={pending || t`Processing transaction...`} />, {
```
parrot9design, line 52: The previous version used a success message instead of a pending one.

If we need to add a success message as well, let's use `Transaction sent to the network.`

PR #941, opened 2022-04-06
add berlin testnet

octocat512:


BennyOptions:
Might not need this PR, consider closing it next week if it proves not useful

Issue #932, opened 2022-03-30
Refactor Ledger Integration


BennyOptions:
More enhancements to be made on Ledger integration

BennyOptions:
merged in PR #943

PR #943, opened 2022-04-09
fix: prevent ledger was deactivated on balanced when user login the same wallet address on another platform

shallcdn:
### Step to reproduce:

1.    Login Balanced by Ledger wallet with address hx….01

2.    Login OMM [App | Omm](https://app.omm.finance/#/home)  with the same wallet address as above hx….01

3. Try to perform a transaction on Balanced,

**Actual**: Nothing happened
**Expectation**: Work with any transaction

### Solution:

Renew transport with any transaction need to confirm on Ledger

#932

### DIFF
--- a/src/packages/BalancedJs/index.ts
+++ b/src/packages/BalancedJs/index.ts
@@ -155,8 +155,9 @@ export class BalancedJs {
   }
 
   async initializeTransport(): Promise<void> {
-    debugger;
-    if (this.contractSettings.ledgerSettings.transport?.device?.opened) {
+    // extract ledger address index from path to create a new transport with the correct wallet address
+    let addressIndex = Number(this.contractSettings.ledgerSettings.path?.replaceAll("'", '').split('/')[4]);
+    if (this.contractSettings.ledgerSettings.transport?.device?.opened && addressIndex >= 0) {
       this.contractSettings.ledgerSettings.transport.close();
       this.contractSettings.ledgerSettings = {} as LedgerSettings;
       const transport = await TransportWebHID.create();
@@ -166,7 +167,7 @@ export class BalancedJs {
       this.inject({
         legerSettings: {
           transport,
-          path: getLedgerAddressPath(-1),
+          path: getLedgerAddressPath(addressIndex),
         },
       });
     }
--- a/src/packages/BalancedJs/index.ts
+++ b/src/packages/BalancedJs/index.ts
@@ -154,7 +154,7 @@ export class BalancedJs {
     this.contractSettings.resetLedgerSettings();
   }
 
-  async initialiseTransport(): Promise<void> {
+  async initializeTransport(): Promise<void> {
     debugger;
     if (this.contractSettings.ledgerSettings.transport?.device?.opened) {
       this.contractSettings.ledgerSettings.transport.close();
--- a/src/packages/BalancedJs/index.ts
+++ b/src/packages/BalancedJs/index.ts
@@ -1,3 +1,4 @@
+import TransportWebHID from '@ledgerhq/hw-transport-webhid';
 import BigNumber from 'bignumber.js';
 import { Converter as IconConverter } from 'icon-sdk-js';
 import { isEmpty } from 'lodash';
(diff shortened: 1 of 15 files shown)


lijac (commented):
Thank you!
I couldn't test with a ledger, does this change have a direct impact on the issue?
`changeShouldLedgerSign(true);` => `await changeShouldLedgerSign(true);`
I'm asking since I haven't seen using `await` when calling a hook before.

shallcdn:
> Thank you! I couldn't test with a ledger, does this change have a direct impact on the issue? `changeShouldLedgerSign(true);` => `await changeShouldLedgerSign(true);` I'm asking since I haven't seen using `await` when calling a hook before.

Hi @lijac, I think it doesn't impact the issue directly. However, when I found the place to add this block of code [ (initialize a new transport for ledger wallet) ](https://github.com/balancednetwork/balanced-network-interface/pull/943/files#diff-3abfa5f69436e97db717daa3bbb76e71e0308c030703d4c4604dd9e85f633d10L49-R54) to fix this issue, I see this place, I can add the codes as above, and don't need to modify a lot.

Do you have any ideas? Please advise me.

src/store/application/hooks.ts
```diff
@@ -46,7 +48,10 @@ export function useCurrentLedgerAddressPage(): AppState['application']['currentL
 export function useChangeShouldLedgerSign(): (shouldLedgerSign: boolean) => void {
```
lijac, line 48: The return value will be `Promise<void>`

lijac (commented):
We can just call `changeShouldLedgerSign(true);` without `await`.

src/packages/BalancedJs/index.ts
```diff
@@ -148,4 +149,22 @@ export class BalancedJs {
   getContract(address: string): IRC2 {
     return new IRC2(this.contractSettings, address);
   }
+
+  async initialiseTransport(): Promise<void> {
+    debugger;
```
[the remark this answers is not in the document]
lijac: Let's use `initializeTransport` following the US style.

shallcdn:
> We can just call `changeShouldLedgerSign(true);` without `await`.

Because function `initializeTransport` contain this block code
```
const transport = await TransportWebHID.create();
```
 So that if we don't use `await` for `changeShouldLedgerSign(true);` the transport of Ledger cannot renew before calling to create a new transaction

lijac:
> So that if we don't use `await` for `changeShouldLedgerSign(true);` the transport of Ledger cannot renew before calling to create a new transaction

Well, do we have to ensure ledger transport renewal is complete before moving to the next operation? Can you check that again with the ledger?

src/packages/BalancedJs/index.ts
```diff
@@ -148,4 +149,21 @@ export class BalancedJs {
...
+      this.contractSettings.ledgerSettings.transport.close();
+      this.contractSettings.ledgerSettings = {} as LedgerSettings;
+      const transport = await TransportWebHID.create();
+      if (transport.setDebugMode) {
+        transport.setDebugMode(false);
+      }
+      this.inject({
+        legerSettings: {
+          transport,
+          path: getLedgerAddressPath(-1),
```
lijac, line 164: Can you leave a comment on what `-1` means?

shallcdn:
> > So that if we don't use `await` for `changeShouldLedgerSign(true);` the transport of Ledger cannot renew before calling to create a new transaction
> 
> Well, do we have to ensure ledger transport renewal is complete before moving to the next operation? Can you check that again with the ledger?

Yes, I have checked it. If I remove `await`, the Iconex/Hanna wallet will pop up instead ledger

BennyOptions:
I tested with ledger and it seems to work @shallcdn

lijac:
Thank you! Let me think about possible improvements.

src/packages/BalancedJs/index.ts
lijac: Updated

shallcdn:
> I tested with ledger and it seems to work @shallcdn

Thank you!

shallcdn:
> Thank you! Let me think about possible improvements.

Thank you

octocat512:
@shallcdn please resolve the conflicts.

lijac (approved):

PR #959, opened 2022-04-24
feat: show wallet button in all sizes

hetfly:


parrot9design:
If I tap the address to copy it, then scroll down the page, the `Copied` message follows:
IMG_441460E4BB10-1

When I try this in a Brave window below 800px, the start of the wallet address is overlapped:
Screen Shot 2022-04-25 at 8 51 32 AM
(can't reproduce on mobile, though 🤔)

Other than that, this is good!

parrot9design (approved):
Looks good, thanks @hetfly!

octocat512:
@hetfly @parrot9design  The `Copied` text is suddenly shifted to the center whenever clicking the address text after opening the menu.

hetfly:
@octocat512 that's intended. But I guess there is enough space even on 320px screens so we can keep it on the left.

PR #976, opened 2022-05-13
Revert "Revert "Feat/multicall""

lijac:
Reverts balancednetwork/balanced-network-interface#974


BennyOptions:
@lijac can you share if you have found the cause of the bugs in this feature? Just a brief update, thanks!

- [ ] sICX/ICX price chart bug
- [ ] Trade page crash bug

lijac:
> * [x]  sICX/ICX price chart bug
> * [ ]  Trade page crash bug

sICX/ICX price bug - bug in refactoring getting queue pair info
Since testing is done in Sejong testnet where sICX/ICX rate is 1, this bug went unnoticed.

Trade page crash bug - cannot reproduce it, skipping for now

PR #978, opened 2022-05-19
Feat/add bbaln interface

hetfly:


BennyOptions:
@parrot9design IMO user should know "at what point does locking extra BALN have 0 impact on my rewards." In the current design, the user needs to guess/test until they find the perfect amount of bBALN needed to reach max boost. It would be good to display "bBALN needed = 300", where 300 would reach max boost for ALL pools (so it just displays the pool that requires the most bBALN to reach max boost"


https://user-images.githubusercontent.com/43685868/195434725-a2ff85a2-bf47-4850-9bbc-8e51b350f4da.mov

BennyOptions:
This will be very small for most people, so think 4 decimals would be a good idea (or maybe come up with another statistic worth displaying, or maybe make more room in the section itself, just throwing ideas). Also, maybe a tooltip that explains it's "my bBALN / total bBALN". The current display could be interpreted as your network fee yield

Screen Shot 2022-10-12 at 3 56 43 PM

BennyOptions:
@hetfly the numbers should update after successful confirmation of the BALN lock. I was screensharing with you when it happened, lmk if you don't remember the issue.

BennyOptions:
Currently, this is the only place to see your actual APY after boost. Users need to first select the pool, then they can see APY. As a user, I'd like to see all my APY's post-boosting for my various pools rather than needing to check each of them one by one. Maybe replace pool share with APY in liquidity details?

Screen Shot 2022-10-12 at 4 07 22 PM

parrot9design:
## Wallet

- [ ] When you view an asset, then click away, you have to reopen the wallet menu to check your other tokens. Might be helpful to keep the wallet active in the background so it's easy to go back to.
- [ ] Add a search icon before the text in the search field (should be in the search fields on the Trade page, as well)
Screen Shot 2022-10-13 at 4 41 30 PM

- [ ] Can we add a footer to the wallet to include `Total` and the USD value?
- [ ] After unstaking BALN from the wallet, the active asset changed to sICX. It should show the BALN Send tab, and ideally remove the staking/unstaking sections completely after all BALN has been unstaked.

---
## Empty state

- [ ] Let's keep the bottom row of the boosted section visible even if you haven't locked up BALN yet. For example, this is the initial state on Omm:

Screen Shot 2022-10-13 at 2 19 13 PM


- [ ] For the values, show `1 x` under market and/or liquidity rewards if you've participated in those activities, otherwise show a `–`.

- [ ] Related to [@BennyOptions' comment](https://github.com/balancednetwork/balanced-network-interface/pull/978#issuecomment-1276669863), we could also change `Network fees` to `bBALN rewards`, and display the same  APY we use on the Stats page.

- [ ] Add a tooltip for `bBALN rewards`:
> The average annual return per bBALN, calculated from the last 30 days of network fees ({$amount}).

- [ ] Let's also update the main `Boost rewards` tooltip to:
>Lock up BALN to hold voting power and boost your earning potential by up to 2.5 x.
>The longer you lock up BALN, the more bBALN (Boosted BALN) you'll receive; the amount will decrease over time.

---

## Migration from staked -> locked BALN
 
- [ ] We should try to prevent users from holding both staked BALN and bBALN. To encourage people to migrate from the staked -> locked model, let's display this UI for anyone who has BALN staked in their wallet:

Screen Shot 2022-10-13 at 5 01 46 PM

```
<p class="text-center"><span class="text-bold">BALN staking has been upgraded to a lock-up model called Boosted BALN.</span><br><span class="label intro-label">Unstake your BALN, then lock it up for 1 week – 4 years. You'll receive bBALN, which determines your voting power, network fee earnings, and BALN rewards.</span></p>
```

Confirmation modal:
Screen Shot 2022-10-13 at 17 07 34

Notification messages:

>Unstaking BALN…

>{amount} BALN added to your wallet.

---

## Locking up BALN

- [ ] Let's change the `Boost` button label to `Lock up BALN`.

Screen Shot 2022-10-13 at 2 20 56 PM

- [ ] Change the date formatting for the confirmation modal:
Screen Shot 2022-10-13 at 5 58 19 PM

--
After locking, it loaded the withdraw view, even though I didn't lock up all BALN and the unlocked date is Oct 20:
Screen Shot 2022-10-13 at 2 25 22 PM

After about a minute, it seemed to fix itself, but it took a few hours until I could lock up additional BALN. Assume this is just because of the work still being done on the early withdrawals?

---

## Maximise your boost

- [ ] Remove the `Max rewards` tooltip, as it's only displayed on loan rewards anyway.

- [ ] Instead, display a tooltip next to the bBALN amount while in edit mode to show how many bBALN are needed to maximise BALN rewards.

> {amount} bBALN required for maximum rewards.

- [ ] When not in edit mode, add it after a line break to the main `Boost rewards` tooltip, similar to how we handled the Rebalancing tooltips.

We'll likely enhance this even further, but this will work for v1. 

---

## Withdraw early


- [ ] Update the warning message to `Pay a 50% fee to unlock your BALN early.`
- [ ] Make the slider red to emphasise the situation:
Screen Shot 2022-10-13 at 6 12 44 PM

- [ ] Update withdraw modal (remove decimals for round numbers; add `all` to the button (`Unlock all BALN for a 50% fee`):
Screen Shot 2022-10-13 at 5 55 55 PM


---
## Supply liquidity section - APY info

- [ ] Show the BALN APY range for each liquidity pool:
Screen Shot 2022-10-13 at 5 51 38 PM


---
## Liquidity pools section

- [ ] In the `Your pools` section, change `Pool share` column to `BALN APY`
- [ ] In the `Incentivised pools` section, update the APY column to show min-max ranges for the `BALN` values (use max value to calculate the order)
Screen Shot 2022-10-13 at 5 52 02 PM

- [ ] Update the APY tooltip:
>The BALN APY is calculated from the USD value of BALN rewards allocated to a pool. Your rate will vary based on the amount of bBALN you hold.<br><br>The fee APY is calculated from the swap fees earned by a pool in the last 30 days.<br><br>Impermanent loss is not factored in.


----

## Misc

- [ ] Make the locked indicator always point into the section, so it doesn't overlap with other elements. The position in this screenshot should be used when the slider is above 50%. Flip it the other way for use below 50%.
Screen Shot 2022-10-13 at 5 36 20 PM


Note: I haven't been able to test the standard withdraw flow yet.

parrot9design:
- [ ] Let's add 10px of padding to the bottom of the heading here:
Screen Shot 2022-10-19 at 6 01 14 PM

- [ ] Let's bold the final sentence in this tooltip:
Screen Shot 2022-10-19 at 6 08 31 PM

- [ ] Update Vote info to display bBALN values instead of BALN (new proposal page, vote confirmation modals, and voting weight)

I haven't tested the full flow on mobile or from initial onboarding, will do that tomorrow.

parrot9design:
The tooltips have a higher z-index than the wallet, so if you open the wallet while in the edit state, the tooltips appear over the top:

Screen Shot 2022-10-25 at 1 50 07 PM

---

Let's update the tooltip content here:
Screen Shot 2022-10-25 at 1 48 37 PM
>Lock up BALN to earn network fees.

---

Screen Shot 2022-10-25 at 1 53 35 PM

- Add comma separation
- Add the unlock date:
>14,250 BALN locked until 7 November, 2022.

---

If you haven't taken out a loan or supplied liquidity to any incentivised pools, the boost value should be `N/A`.
Screen Shot 2022-10-25 at 2 37 17 PM

---

Remove the label style from the second paragraph in this tooltip, and add a horizontal separator before the max bBALN info. We could increase the width a little to make the text wrap better.
Screen Shot 2022-10-25 at 9 38 03 AM

src/app/components/Form/index.tsx
```diff
@@ -65,7 +65,7 @@ const NumberInput = styled.input`
 
 const CurrencyUnit = styled.span``;
 
-const inputRegex = RegExp(`^\\d+(?:\\\\[.])?\\d*$`);
+export const inputRegex = RegExp(`^\\d+(?:\\\\[.])?\\d*$`);
```
shallcdn, line 68: Please give a brief comment to explain it in code, thanks

shallcdn:
Screen Shot 2022-10-27 at 23 21 10
It may break UI in mobile

shallcdn:
Screen Shot 2022-10-27 at 23 33 02

I'm unsure whether this case is correct; the staking feature on the popup differs from the boost rewards feature. Once I try to stake BALN, I see the Boost Rewards appear the message is like the attached image, although I already locked up BALN to boost my earning potential.
Screen Shot 2022-10-27 at 23 28 19

It makes me a bit confused

shallcdn (approved):
Please take a look my comments, the rest LGTM

parrot9design:
@shallcdn, the BALN staking tab in the wallet is there to test the initial migration. It'll be removed when the feature goes live so users can't stake BALN again.

---

@hetfly have some more feedback after testing the withdrawal flow.


When BALN is ready to withdraw, the liquidity rewards are showing as above 1.0 x, even though the UI says I have no bBALN:
Screen Shot 2022-10-28 at 4 14 31 PM

We could also change the network fee value to `N/A` whenever there's no BALN locked to make it more obvious when you're not earning any.

----

Let's update the confirmation modal:
From this | To this
--------- | ---------
Screen Shot 2022-10-28 at 4 19 53 PM | Screen Shot 2022-10-28 at 4 36 35 PM

>Withdraw Balance Tokens?
>
>**200 BALN**
>
>You'll need to relock BALN to hold voting power and boost your earning potential.
>
>Cancel | Withdraw BALN

---

Let's add `Available since {date}` where the date is in other states:
Screen Shot 2022-10-28 at 4 41 11 PM

Speaking of dates... are they displayed in UTC time, or your local time? Seems to be UTC time based on testing, but local time would be a better experience, especially for users ahead of that time zone.

---

When unlocking BALN early, let's adjust the banner messages to use "unlock" instead of "withdraw":

>Unlocking BALN early...

>Unlocked {amount} BALN.

---

On mobile, the `Liquidity rewards` tooltip is difficult to trigger, and when it does, it's broken like Shall mentioned. 

---

Can't test this on Berlin, but will these transactions populate the Activity History section?

src/app/components/Form/index.tsx
hetfly: @shallcdn it was an old copy paste code from earlier. Refactored now, thanks for pointing that out!

hetfly:
@parrot9design 
> When BALN is ready to withdraw, the liquidity rewards are showing as above 1.0 x, even though the UI says I have no bBALN

These are the actual data from smart contract. To refresh them user must claim rewards or withdraw.

> Speaking of dates... are they displayed in UTC time, or your local time? Seems to be UTC time based on testing, but local time would be a better experience, especially for users ahead of that time zone.

UI shows local time. I think I mentioned the wrong time to you regarding the lock checking. It's 00:00 UTC every Thursday so the issue would be in time zones that are behind. See the video and check how the lock changes the date for Pacific time.

https://user-images.githubusercontent.com/13098459/198892957-dd5697e0-5a3f-4825-bc3d-54677ad3d7d2.mov

> Can't test this on Berlin, but will these transactions populate the Activity History section?

Not from the launch. Withdrawals shouldn't be a problem but the BALN lock is a simple token transfer. Will look into it after I lock on mainnet. Depending on the overall situation after the release, this could be done in an hour or so. 

The rest from the feedback is fixed.

parrot9design:
When the browser width is 500-508 pixels, the liquidity pool labels are misaligned. Let's push them underneath:
Screen Shot 2022-10-31 at 7 48 06 PM

---
At mobile, you can tap the wallet address in the dropdown but it doesn't do anything.

We should make it possible to copy the address, add a copy icon at the end, and include an `Address copied` message similar to at desktop.

---
If you don't qualify for loan or liquidity rewards the concept of "maximum BALN" doesn't apply, so we could hide this message from those users completely:

parrot9design (approved):
Left a couple comments, but otherwise this looks good to merge. 💪

PR #982, opened 2022-05-27
feat/add stability fund UI

### LINKED ISSUE #975: UX around interacting with the Stability Fund
It would be good to allow users to interact with the Stability Fund directly. Let's think about the best way to do this. The functionality allows users to swap whitelisted assets (currently IUSDC and USDS) for bnUSD for 1:1 minus a 0.5% fee. Depending on how much is left in the stability fund, the users can also swap bnUSD back for any of the whitelisted assets 1:1 minus a 0.5% fee.

hetfly:
close #975


parrot9design:
Great work so far! A few comments:

- Let's show the Stability Fund option whenever a number is entered for a relevant swap – people shouldn't need to hold the assets in their wallet or be signed in to assess potential trade opportunities.

- As these are stablecoins, let's limit the numbers to 2 decimal places.

- If the Stability Fund isn't available for a certain trade (or is within ~$2-5 of the limit) we should hide the option so it doesn't become a nuisance. (Open to suggestions as to what the lowest useful trade size would be.)

- When the amount you enter is over the maximum available in the fund, let's keep the `OR` and divider in the tooltip, and use this content:

> Use the Stability Fund to swap a maximum of [2 IUSDC]() for bnUSD.

hetfly:
Thanks for the feedback, code is updated.
At the moment, there is a cushion of 5$ for swaps `whitelisted token -> bnUSD` since it's roughly the amount when transactions stopped randomly failing with error `reverted(0)`

parrot9design:
This is getting close now. Just a couple things:

- The content is a little small – I think the link style is overriding it? Let's make sure it's never less than 14px.

- Let's add the `?` tooltip to the max available content:
Screen Shot 2022-05-31 at 2 04 49 PM

- Right now it's actually cheaper to swap USDS for bnUSD than to go through the Stability Fund. 😅 This is probably a terrible idea, but @BennyOptions what are your thoughts on only displaying the fund when it's beneficial to use it? @hetfly, would that be difficult to handle from a logic/performance perspective?

- I traded some IUSDC for USDS, then swapped USDS for bnUSD through the Stability Fund. The activity history section shows two transactions, but the info isn't correct, and both link to this transaction (not sure how they were lumped together?): https://tracker.icon.foundation/transaction/0x40063f62991dbc157b894a51622727a55b129c81ffdd770337ccac67e8cd24db 
Screen Shot 2022-05-31 at 2 19 32 PM

hetfly:
Good spotting with the font size.

Tooltip added.

> * Right now it's actually cheaper to swap USDS for bnUSD than to go through the Stability Fund. 

It won't be a problem to implement but for what it's worth I would leave it there even if the stability fund swap is a worse trade. Helps with performance and it shows only for the eligible pairs anyway and avoids thinking that stability fund isn't suddenly available. Also it could be better to trade at lower amounts and swap at higher amounts because of the slippage. 

I can't help with the tx bundling but to show Stability fund usage in history, BE needs to add another type to this endpoint
https://balanced.sudoblock.io/api/v1/docs#/stats/get_stats_transaction_history_api_v1_stats_transaction_history_get
after that it's just a few lines of code on FE.

BennyOptions:
@hetfly it gets cutoff for me for some reason. Using Chrome, not full screen, not sharing screen.

Screen Shot 2022-05-31 at 9 30 39 PM

src/app/components/StabilityFund/index.tsx
```diff
@@ -0,0 +1,255 @@
...
+              <span onClick={() => (account ? setShowFundSwapConfirm(true) : toggleWalletModal())}>
+                {t`Use the Stability Fund to swap ${`${new BigNumber(sendAmount.toSignificant()).toFormat(
+                  2,
+                )} ${sendSymbol}`} for ${`${new BigNumber(sendAmount.subtract(feeAmount).toSignificant()).toFormat(
+                  2,
+                )} ${receivedCurrency?.symbol}`}`}
+              </span>
+
+              <QuestionHelper
+                text={t`The Stability Fund allows you to mint or burn bnUSD 1:1 for approved stablecoins.`}
```
parrot9design, line 169:
```suggestion
                text={t`The Stability Fund lets you mint or burn bnUSD 1:1 for approved stablecoins, minus a 0.5% fee.`}
```

I think it's better to be specific about the fee %, rather than just say "minus a small fee", but if the community can vote to adjust the fee, we'll need to add an API call so a UI update isn't required to show accurate info.

hetfly: updated

hetfly:
@BennyOptions check it out, should work properly now.

In this setup, where Fund UI is never cutoff, this could happen 
Screenshot 2022-06-01 at 17 13 32
but there is always enough room to scroll down so that the tooltips shows below the button as intended. 

It remains forced to appear below the button all the time on mobile.

parrot9design (approved):
I can't check to see how this is handled in Chrome (Brave only here), but aside from that, this should be good to merge.

In the last round of testing, none of the Stability Fund transactions showed up in the activity history section, but this is something we can address in a month or so when we can pull this data from the ICON node API instead.

octocat512:
@hetfly Please resolve the conflicts.

src/app/components/trade/SwapPanel.tsx
```diff
@@ -217,6 +225,17 @@ export default function SwapPanel() {
 
   const hasEnoughICX = useHasEnoughICX();
 
+  const SwapButton = () =>
```
octocat512, line 228: use a variable element not a component. and `useCallback`.

octocat512 (changes requested):

src/store/stabilityFund/hooks.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (currencies.INPUT?.isToken && currencies.OUTPUT?.isToken) {
+    isEligible =
+      (whitelistedTokenAddresses.indexOf(currencies.INPUT.address) >= 0 &&
+        currencies.OUTPUT.address === bnUSDAddress) ||
+      (whitelistedTokenAddresses.indexOf(currencies.OUTPUT.address) >= 0 && currencies.INPUT.address === bnUSDAddress);
+  }
+  return isEligible;
+}
+
+export function useMaxSwapSize() {
```
octocat512, line 77: return data type: CurrencyAmount<Token>

octocat512 (changes requested):

src/store/stabilityFund/hooks.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+    });
+
+    const data: string[] = await bnJs.Multicall.getAggregateData(cds);
+
+    const balances: { [key: string]: CurrencyAmount<Token> } = {};
+    data.forEach((balance, index) => {
+      const address = whitelistedTokens[index];
+      const token = SUPPORTED_TOKENS_LIST.filter(token => token.address === address)[0];
+      balances[address] = toCurrencyAmountFromRawBN(token, new BigNumber(balance));
```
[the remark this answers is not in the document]
lijac: `balance` might be not an integer?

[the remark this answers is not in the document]
hetfly: `balance` is obtained straight from the contract as hex number in this case

src/app/components/StabilityFund/index.tsx
```diff
@@ -89,7 +89,9 @@ const StabilityFund = ({ clearSwapInputOutput, setInput }: StabilityFundProps) =
   const receivedCurrency = trade?.outputAmount.currency;
   const isBnUSDGoingIn = sendSymbol === 'bnUSD';
   const hasFundEnoughBalance =
-    sendAmount && maxSwapSize ? new BigNumber(sendAmount.toFixed(4)).isLessThanOrEqualTo(maxSwapSize) : 'loading';
+    sendAmount && maxSwapSize
+      ? new BigNumber(sendAmount.toFixed()).isLessThanOrEqualTo(new BigNumber(maxSwapSize.toFixed()))
```
octocat512, line 93: You can use compare functions in the `fraction`  type.

src/store/stabilityFund/hooks.ts
```diff
@@ -2,16 +2,15 @@ import { CallData } from '@balancednetwork/balanced-js';
 import { CurrencyAmount, Fraction, Token } from '@balancednetwork/sdk-core';
 import BigNumber from 'bignumber.js';
```
octocat512, line 3: remove BigNumber import. don't BigNumber AMAP.

src/store/stabilityFund/hooks.ts
```diff
@@ -74,73 +73,76 @@ export function useIsSwapEligible(): boolean {
...
     .div(toDec.valueOf())
     .div(toPercent.valueOf())
     .toNumber();
 }
 
-export function useFeeOut() {
-  const { data } = useBnJsContractQuery<number>('StabilityFund', 'getFeeOut', [], false);
+export function useFeeOut(): number {
+  const { data } = useBnJsContractQuery<string>('StabilityFund', 'getFeeOut', [], false);
   return IconConverter.toBigNumber(data || 0)
```
octocat512, line 146: use `parseUnits` or `formatUnits` in utilis

src/store/stabilityFund/hooks.ts
```diff
@@ -74,73 +73,76 @@ export function useIsSwapEligible(): boolean {
...
+      return limits;
+    },
+  );
 }
 
-export function useFeeIn() {
-  const { data } = useBnJsContractQuery<number>('StabilityFund', 'getFeeIn', [], false);
+export function useFeeIn(): number {
+  const { data } = useBnJsContractQuery<string>('StabilityFund', 'getFeeIn', [], false);
   return IconConverter.toBigNumber(data || 0)
```
octocat512, line 138: use `parseUnits` or `formatUnits` in utilis

octocat512 (changes requested):

BennyOptions:
@octocat512 please merge if requested changes are fixed.

octocat512 (approved):

PR #983, opened 2022-05-31
feat: add user logout after an hour

### LINKED ISSUE #973: Automatically sign users out after 60 minutes
Now that we have the ability to keep users signed in on page refresh, we should limit how long that session lasts. It's quite disorienting to load Balanced hours (or days) after the last visit and still be signed in.

30-60 minutes should be long enough.

Related: https://github.com/balancednetwork/balanced-network-interface/pull/960

hetfly:
close #973


BennyOptions:
@octocat512 should be good to merge

PR #987, opened 2022-06-01
refactor/use balancednetwork packages

octocat512:


RiseBlock:
Screen Shot 2022-06-10 at 10 21 13

Blank page showing this error, this happens with a wallet having assets. Looks like empty wallet doesn't encounter this issue.

PR #998, opened 2022-06-08
feat: add vote page enhancements

### LINKED ISSUE #991: Fix issues on the Vote page
- The indicator to alert users to a new vote no longer appears
- The start of the Approve/Reject bars looks funky in Brave (normal in Firefox & Safari, not sure about Chrome):
Screen Shot 2022-06-06 at 12 49 48 PM
Screen Shot 2022-06-06 at 12 50 26 PM

- All votes that have been approved but not enacted via smart contract now say `Voting has ended`. They should say `Approved`.

Another issue that can be tackled separately if necessary: if you're signed in and viewing past votes that you participated in, we should show whether you approved or rejected them.

hetfly:
close #991

### DIFF
--- a/src/app/components/AppBar/index.tsx
+++ b/src/app/components/AppBar/index.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { Trans } from '@lingui/macro';
+import BigNumber from 'bignumber.js';
 import { NavLink } from 'react-router-dom';
 import { Text } from 'rebass/styled-components';
 import styled from 'styled-components';
@@ -9,6 +10,7 @@ import { ReactComponent as HomeIcon } from 'assets/icons/home.svg';
 import { ReactComponent as TradeIcon } from 'assets/icons/trade.svg';
 import { ReactComponent as VoteIcon } from 'assets/icons/vote.svg';
 import { useActiveProposals } from 'queries/vote';
+import { useBALNDetails } from 'store/wallet/hooks';
 
 import { notificationCSS } from '../home/wallets/utils';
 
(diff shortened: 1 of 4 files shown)


dbrehmer:
@hetfly - I see the `&& stakedBALN.isGreaterThan(0)`, which should fix my concern. Thanks!

parrot9design:
Thanks @hetfly! Apologies, I wasn't very clear in the issue but I don't think we need to show it at the top level. I was thinking we could display it the same way we do while the vote is active, just without the `Change vote` option:
Screen Shot 2022-06-09 at 4 55 11 PM

hetfly:
@parrot9design oh yea, that makes much more sense, done.

Regarding the visual bugs of the voting bars, I've removed the animation. It wasn't that impactful anyway IMO and was causing this. This is the 3rd time I think we cared for this, not worth it I guess.

@dbrehmer sorry if spam but in case you missed it - Approved proposals contain status of `Active` instead of `Succeeded`. Could you please check that out? Those are the proposals that currently display `Voting has ended` 
Screenshot 2022-06-09 at 15 41 49

parrot9design:
Aside from the Approved proposals bug, this is looking great! Love that it shows the voting weight at the time of the proposal, as well. 👏

parrot9design:
@hetfly, I believe the only thing outstanding on this PR is the vote status?

@dbrehmer All approved (but not enacted) proposals currently display `Voting has ended` because the data contains an `Active` status instead of `Succeeded`. Could you please take a look when you have the chance?

PR #1006, opened 2022-06-20
Feat/add price info to currency selector

### LINKED ISSUE #979: Add price info to the Swap asset selectors
A recent discussion on Discord revealed that some community members rely on the ICX & bnUSD price in the position details section to decide when it's a good time to swap assets. Some even take out a tiny loan in order to see this info.

We can do better.

We should add the price to the Swap asset selectors so people don't have to jump back and forward between the Home and Trade pages. 

**Signed out** | **Signed in**
-------------- | ---------------
 | price-signed-in

As we already have this data on the Stats page, assuming it should be no issue to display it here, as well.

hetfly:
close #979

### DIFF
--- a/src/app/components/SearchModal/CurrencyList.tsx
+++ b/src/app/components/SearchModal/CurrencyList.tsx
@@ -1,19 +1,22 @@
 import React, { useEffect, CSSProperties, useState, useCallback } from 'react';
 
-import { Currency, Token } from '@balancednetwork/sdk-core';
+import { Currency, Fraction, Token } from '@balancednetwork/sdk-core';
 import { Trans } from '@lingui/macro';
 import BigNumber from 'bignumber.js';
-import { isMobile } from 'react-device-detect';
+import { isIOS, isMobile } from 'react-device-detect';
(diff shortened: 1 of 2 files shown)


parrot9design (changes requested):
This is excellent! There's just one minor thing to fix: we should add commas to the $ value of assets in your wallet to make them easier to read. So in this example, it should be `$1,721.76`:
Screen Shot 2022-06-21 at 10 30 38 AM

Also, wondering if we should include more decimal places for assets like METX, to avoid displaying a $ value of 0?
Screen Shot 2022-06-21 at 10 42 15 AM

parrot9design:
That's looking much better, thanks @hetfly! Let's remove the bold style from the token price when signed in so that it's the same weight as the $ value of your wallet holdings, then it should be good to merge.

Screen Shot 2022-06-22 at 10 54 16 AM

hetfly:
@parrot9design done

parrot9design (approved):

parrot9design:
Perfect, thanks @hetfly!

PR #1018, opened 2022-06-27
fix:update wrong decimals of usdt/usdc of stakedlp

shallcdn:
Fix wrong decimals to calculate value of usdt/usdc for stakedlp


BennyOptions:
Changes work

Issue #1050, opened 2022-07-28
Fix base/quote currency display


BennyOptions:
This pair is backwards. bnUSD should NEVER be on the left for any pair.

sICX/BALN is also incorrect. It should be BALN/sICX.

<img width="232" alt="Screen Shot 2022-07-28 at 3 11 24 PM" src="https://user-images.githubusercontent.com/43685868/181618912-302a86d0-34e2-4b29-be62-9276e5333317.png">

molly-tran:
Please check help me !

PR #1054, opened 2022-08-05
fix-currency-display

molly-tran:


src/app/components/trade/LiquidityDetails.tsx
```diff
@@ -425,7 +425,12 @@ const PoolRecord = ({
     <>
       <ListItem>
         <StyledDataText>
-          <DataText>{`${aBalance.currency.symbol || '...'} / ${bBalance.currency.symbol || '...'}`}</DataText>
+          {aBalance.currency.symbol == 'bnUSD' ||
```
shallcdn, line 428: Should use '===' instead of  '=='

shallcdn (changes requested):
Please update as my comment

shallcdn (approved):
LGTM

BennyOptions:
Hey @molly-tran @shallcdn this should not be a hard-coded solution. There is always a base_currency and quote_currency. Base is on the left, quote is on the right. Something is incorrect in the logic of the app, simple hard coding is not the ideal solution.

BennyOptions:
Also, can you guys set this PR to main net so I can test with my wallets that have liquidity? Otherwise I need to get more test tokens which will take more time.

shallcdn:
> Also, can you guys set this PR to main net so I can test with my wallets that have liquidity? Otherwise I need to get more test tokens which will take more time.

@BennyOptions I have deployed the current change with prod environment, please take a look!

BennyOptions:
> > Hey @molly-tran @shallcdn this should not be a hard-coded solution. There is always a base_currency and quote_currency. Base is on the left, quote is on the right. Something is incorrect in the logic of the app, simple hard coding is not the ideal solution, because it will display new pools incorrectly.
> 
> Hi @BennyOptions Where can I check the correct order of the pool pairs? on this page, https://stats.balanced.network/ Am I correct?

Yes that's correct.

Also, as a rule, sICX is always on the right side (except for sICX/bnUSD). bnUSD is also always on the right side (never any exceptions)

BennyOptions:
The "your supply" column is still backwards.

Screen Shot 2022-08-10 at 2 57 27 PM

BennyOptions:
One thing that confuses me, @shallcdn do you know why sICX/bnUSD is correct, but all others are incorrect?

shallcdn:
> One thing that confuses me, @shallcdn do you know why sICX/bnUSD is correct, but all others are incorrect?
@BennyOptions 
I have investigated this issue, there is a logic to checking and comparing between two addresses of the base token and quote token to create a pair, 

Eg: Suppose that we have sICX and BALN
sICX address (): 'abd' 
BALN address: 'abc'

The logic will compare 'abd' < 'abc', it will return `false` and create a pair is BALN/sICX, if `true` the pair will be sICX/BALN. 

[JavaScript uses the so-called "dictionary" or "lexicographical" order.](https://javascript.info/comparison#string-comparison) that's why 'abd' is greater than 'abc'

I'm not sure why we used this logic, @octocat512 @hetfly. Do you know why we use this logic to create a Pair?

shallcdn:
PR https://github.com/balancednetwork/balanced-network-interface/pull/1061 fixed issue #1050 so I close this PR

PR #1070, opened 2022-08-29
Feat/add new collateral proposal type

hetfly:


parrot9design:
Let's simplify the DEX content for this tooltip (in the new proposal state only):
Screen Shot 2022-08-30 at 3 24 14 PM
>**DEX:** Use the price from Balanced.
>Requires a bnUSD liquidity pool for this collateral type.

---

The Debt Ceiling tooltip is the same as the Borrow LTV (in the new proposal state only). It should be:
>The maximum amount of bnUSD that can be minted with this collateral type.

---

Let's align the right-hand items at smaller screen sizes:
IMG_0403

---

I keep getting errors in Hana and MyIconWallet when I go to submit the proposal. Does it work in ICONex without any issues?

parrot9design:
I was able to submit a vote successfully (https://balanced-network-interface-8amvt39qy-lydia-labs.vercel.app/vote/proposal/18), but the confirmation modal didn't disappear, even after the notification messages said it had been submitted:
Screen Shot 2022-08-31 at 5 19 13 PM

It wasn't an issue the second time, though. 🤔

Thoughts on redirecting people to the new proposal after it's been submitted? Feels way more useful than looking at an empty proposal page and having to navigate back to find your proposal.

hetfly:
@parrot9design updated

Re the redirect - seems like too much effort since we would have to wait for the creation, fetch the data, redirect based on the fetched ID. I've added redirect after `tx` submit to `/vote/` page where the proposal appears after it's created.

parrot9design:
Not sure why this is happening, but I keep getting a white screen if I try to click through to a few of the proposals that are currently active:

https://balanced-network-interface-pz04yjhgc-balanced-dao.vercel.app/vote/proposal/22
https://balanced-network-interface-pz04yjhgc-balanced-dao.vercel.app/vote/proposal/21
https://balanced-network-interface-pz04yjhgc-balanced-dao.vercel.app/vote/proposal/20

I also keep getting wallet errors when I try to submit a proposal for a new collateral type (persisted even with ICONex). See https://berlin.tracker.solidwallet.io/transaction/0xa5799cad434d9cb48cde74e76021d6d0e5969880f0f60460ac25a95e565ced06 for an example.

For the redirect, happy to jump back to the Vote page, but is it possible to wait until the proposal has been successfully submitted? I had one fail with ICONex after I'd been redirected, would've liked to update some of the values to see if I could get it to go through successfully. Because the transaction hasn't been completed when you're directed to the Vote page, you also don't see your new proposal right away so you still have to wait for that data to load.

If that's a bit awkward, we could always leave the confirmation modal up until the proposal has been submitted, and change the content in the button (like the supply liquidity flow):


Screen Shot 2022-09-01 at 12 02 23 PM

hetfly:
@parrot9design updated

> I also keep getting wallet errors when I try to submit a proposal for a new collateral type

This particular transaction fails because there already is IUSDC collateral added on Berlin. Other issues might be not existing bnUSD pool with collateral type or Band not supporting the symbol provided.

> is it possible to wait until the proposal has been successfully submitted? 

Yea, the redirect now happens upon creation.

parrot9design (approved):

BennyOptions:
@RiseBlock please review this as a priority, thanks!

shallcdn:
The code LGTM but the transaction always returns failed transactions

https://berlin.tracker.solidwallet.io/transaction/0xd79e9ad88386c49c20523cd7c0f3175919b8958282837cd3a2d5f0da120183e3
https://berlin.tracker.solidwallet.io/transaction/0xbdfd4a76e2c81b6e9856be220322fa6c1c5bc68d1715d3f51744de31cb388269

@hetfly Can you take a look at these transactions?

shallcdn (commented):
The code LGTM but the transaction always returns failed transactions

https://berlin.tracker.solidwallet.io/transaction/0xd79e9ad88386c49c20523cd7c0f3175919b8958282837cd3a2d5f0da120183e3
https://berlin.tracker.solidwallet.io/transaction/0xbdfd4a76e2c81b6e9856be220322fa6c1c5bc68d1715d3f51744de31cb388269

@hetfly Can you take a look at these transactions?

AntonAndell:
> The code LGTM but the transaction always returns failed transactions
> 
> https://berlin.tracker.solidwallet.io/transaction/0xd79e9ad88386c49c20523cd7c0f3175919b8958282837cd3a2d5f0da120183e3 https://berlin.tracker.solidwallet.io/transaction/0xbdfd4a76e2c81b6e9856be220322fa6c1c5bc68d1715d3f51744de31cb388269
> 
> @hetfly Can you take a look at these transactions?

ETH is not a supported asset by the underlying oracle. Which in this case would be a dummy oracle i think. I can update to one with dummy prices for ETH and BTC

hetfly:
@AntonAndell thanks!
@shallcdn after Andells update I was able to create a proposal with ETH. Do you still have any issues?

PR #1071, opened 2022-08-29
Update tokenList with Claw token

clawsdev:


hetfly:
taken care of in https://github.com/balancednetwork/balanced-network-interface/pull/1072

PR #1081, opened 2022-09-09
feat: show baln rewards for all positions

hetfly:


parrot9design (approved):
Much better, thanks @hetfly! 🙏

PR #1105, opened 2022-11-02
fix: fix conflicting max rewards info

### LINKED ISSUE #1099: Locking BALN: Conflicting info on max rewards
See below. It says that ~59k bBALN is needed to max rewards, but this account is over 60k and still not at max rewards, one of them is 2.49x.

Screen Shot 2022-10-31 at 3 13 17 PM

hetfly:
close #1099


BennyOptions (approved):
lgtm

PR #1117, opened 2022-11-24
Add live voting

hetfly:


parrot9design:
Love the way you've handled the top section in the signed-out state. 👏💪

When you're signed in and either don't hold BALN or have none locked up, you're presented with options that should be limited to bBALN holders:
Screen Shot 2022-11-30 at 10 15 33 AM

- [ ] Let's make the `Voting power` heading an h2.

- [ ] If you don't have any BALN, the voting power section should show the same message as the Rewards section:
Screen Shot 2022-11-30 at 10 18 41 AM

- [ ] If you have BALN but it's not locked up yet, it should look like this:
198190914-4309ebf6-d4c8-4c2e-989d-ae69b245f5e3

- [ ] The `New proposal` button, your available voting power, and the `Your allocation` column should be hidden until you have bBALN.

---

- [ ] Let's add a tooltip to the `Current allocation` column header, using this content to show the countdown:

Screen Shot 2022-11-30 at 10 49 54 AM

---

Love that you can click to fill the available voting power value for the current pool.
- [ ] Let's make it so you can't click the value until you're in the edit state, and when you are, let's style the value the same as the `Add` text, so it's green with an underline on hover.

Screen Shot 2022-11-30 at 1 33 49 PM

---

- [ ] At the moment the pools are listed alphabetically, but we could add column sorting so you can also view the pools based on allocation/incentives.

---

- [ ] The sentence in the confirmation modal is missing a `.`:
Screen Shot 2022-11-30 at 1 32 01 PM

---

- [ ] The Locked indicator isn't aligned with the slider handle when editing on either page:
Screen Shot 2022-11-30 at 1 39 47 PM

---

Can we move the `Daily incentives` column over a little on mobile without making it too cluttered? It's easy to miss when you're in a signed out state.

hetfly:
@parrot9design 

>The Locked indicator isn't aligned with the slider handle when editing on either page:

Wasn't able to replicate the issue. Tried very low amounts of BALN as well. Is this still happening to you?


>At the moment the pools are listed alphabetically, but we could add column sorting so you can also view the pools based on allocation/incentives.

For signed in user it's listed by users highest allocation. To allow sorting by all of the columns is a bit more complicated in this case, because it mixes multiple data sources. We can't simply reuse the sorting mechanism from pools on the trade page, so I'd suggest revisiting this when there will be more items to vote on.

The rest is fixed.

parrot9design:
>Wasn't able to replicate the issue. Tried very low amounts of BALN as well. Is this still happening to you?

Seems to be an issue with _very small_ amounts. With 6/16 BALN locked I could always reproduce it, but as soon as I claimed rewards (14 BALN), it realigned itself.

---

The columns could still be optimised a little for mobile. The daily incentives column isn't noticeable when signed in or out (at least on an iPhone 13 Pro), and when signed in, the current allocation header spans 3 lines.

 


Otherwise, this looks ready for a code review. 💪

hetfly:
@shallcdn commented/duplicated code from commit above is temporary solution to care for rewards distribution method, that will be deprecated on December 9th.

hetfly:
@parrot9design  I wasn't able to reproduce the slider issue even with those low numbers. I don't think many people will use the interface with less than 20 BALN, probably not worth chasing this.

parrot9design (approved):
This looks ready to merge. The mobile UI is fantastic, really nicely done, @hetfly! 💪 👏

shallcdn (approved):
LGTM

hetfly:
@shallcdn 
> should we update the value of the remaining BALN to correspond with the value of the slide bar once users adjust it?

Do you mean the `0` while user slides below the `Locked` threshold? The zero is there intentionally so that it's clear the whole amount will be unlocked, while the slider is still allowed to be positioned along the way to ensure smooth adjustment going towards the `Locked` threshold as well. Above the locked threshold, the BALN amount corresponds to the slider position.

If are unable to test this on Lisbon I can send you some BALN there, let me know if you want some.

parrot9design:
Just spotted a space missing before the timer in this tooltip:
Screen Shot 2022-12-07 at 8 34 29 AM

For the successful banner message, let's add `of` to make it read better: `20% of bBALN allocated to sICX/ICX.`

hetfly:
@shallcdn sent

PR #1140, opened 2022-12-22
Add slippage warning

### LINKED ISSUE #1139: Color change for slippage percentage
Would be good to incorporate a sliding gradient of color - darkest red being closest to 100%, green being closest to 0. Should be some skew, as 0-1% is green, but anything over 1% should start changing color rapidly. 10% slippage could be dark red instead of 100%. These are just example numbers. Maybe Uniswap [repo](https://github.com/Uniswap/interface) has a benchmark. If you can find what Uniswap uses, please use that.

Screen Shot 2022-12-21 at 12 03 58 AM

Screen Shot 2022-12-21 at 12 06 57 AM

hetfly:
close #1139

### DIFF
--- a/src/app/components/trade/SwapPanel.tsx
+++ b/src/app/components/trade/SwapPanel.tsx
@@ -21,6 +21,7 @@ import SlippageSetting from 'app/components/SlippageSetting';
 import { Typography } from 'app/theme';
 import { ReactComponent as FlipIcon } from 'assets/icons/flip.svg';
 import bnJs from 'bnJs';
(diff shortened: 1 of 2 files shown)


hetfly:
Threshold is set to 2.5%, let me know the ideal one if this is one is bad.

parrot9design (approved):
Nicely done! I'm happy with 2.5%, what do you think @BennyOptions?

BennyOptions:
2.5% is good

PR #1142, opened 2022-12-22
Add BTCB to quotes

### LINKED ISSUE #1134: BTC as quote currency
Need to add BTCB to the list of allowable quote currencies. 

Screen Shot 2022-12-20 at 4 01 05 PM

hetfly:
close #1134


parrot9design:
Works well, although my BTCB balance was too small for the pool to appear below. 😅

The only thing we could improve is the decimal precision on the wallet balance: 
Screen Shot 2022-12-23 at 7 47 24 PM
It should use the value that's shown in the asset dropdown.

PR #1145, opened 2023-01-10
Update live voting UI

hetfly:


parrot9design:
This looks good to merge, thanks @hetfly!

Just wanted to confirm the timer for the daily incentives adjustment – currently it updates at 0:00 UTC on Thursday, but if I remember correctly, the pools won't reward anyone based on the new ratios until 5pm UTC?

hetfly:
> currently it updates at 0:00 UTC on Thursday, but if I remember correctly, the pools won't reward anyone based on the new ratios until 5pm UTC?

@parrot9design this initial delay was set to help with the feature launch. Now the rewards are updated as soon as new allocation is set.

AntonAndell:
> > currently it updates at 0:00 UTC on Thursday, but if I remember correctly, the pools won't reward anyone based on the new ratios until 5pm UTC?
> 
> @parrot9design this initial delay was set to help with the feature launch. Now the rewards are updated as soon as new allocation is set.

No, actually it is part of the design difference between how time in balanced is managed. Due to the design of ve Tokens they update on  Thursday 00.00 UTC. But in balanced our day change is somewhere around 18:00 UTC. So each day at Balanced day change the inflation is minted and distributed to the different sources. This is why there is a "delay". But i think it is fine updating the inflation percentages at 00.00 UTC.

PR #1159, opened 2023-02-01
Update live voting wording

### LINKED ISSUE #1158: Update Liquidity Incentives wording
The current column headers and layout for the Liquidity Incentives section aren't as clear as they could be. We should also show people how much BALN each pool can expect to earn in the future based on the current allocation of votes.

liquidity-incentives

One other issue we can address while we're here:

When people lock up more BALN or increase their lock-up time, their increased bBALN isn't automatically applied to their allocation. If people have allocated votes to at least one pool, we should add a message to the `Lock up Balance Tokens?` confirmation modal to make this situation clear:

Screenshot 2023-02-01 at 6 03 43 PM

>To use your increased voting power to influence liquidity incentives, update your allocation.

hetfly:
close #1158

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@balancednetwork/balanced-js": "1.2.6",
+    "@balancednetwork/balanced-js": "1.2.8",
     "@balancednetwork/hw-app-icx": "1.0.6",
     "@balancednetwork/sdk-core": "1.0.1",
     "@balancednetwork/v1-sdk": "1.0.2",
--- a/src/app/components/home/BBaln/BBalnSlider.tsx
+++ b/src/app/components/home/BBaln/BBalnSlider.tsx
@@ -1,6 +1,7 @@
 import React, { useMemo, useRef, useState } from 'react';
 
 import { addresses } from '@balancednetwork/balanced-js';
+import { Fraction } from '@balancednetwork/sdk-core';
 import { t, Trans } from '@lingui/macro';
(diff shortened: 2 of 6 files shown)


parrot9design:
Slight wording adjustment, sorry!

Screenshot 2023-02-02 at 6 54 48 PM

>To apply your increased voting power to liquidity incentives, update your allocation for each pool.


The message seems to appear for all bBALN adjustments. We should set it to only appear when there's less than 100% voting power available, as it's not relevant to people who haven't allocated any votes yet.

Also, in this example it shows 0 BALN before and after. This will be based on the wallet balance, but I think the BALN locked before/after would make more sense?

hetfly:
@parrot9design Good point with the message showing relative to already used voting power.

You might be right with the BALN before/after in the modal as well. Not sure, though, if I like that the `Lockup Balance Tokens` and `after` amounts will be always the same. It's updated, have a look and let me know. I'm fine with both options.

>Most of the pools show 0 BALN in the future incentives column. Is that expected?

No, it was a bug happening after reallocation. Fixed now.

parrot9design:
>You might be right with the BALN before/after in the modal as well. Not sure, though, if I like that the Lockup Balance Tokens and after amounts will be always the same. It's updated, have a look and let me know. I'm fine with both options.

Hmm, you make a good point there. Not too fond of the repetition either, but I think the new behaviour makes sense as the Collateral and Loan modals also use their respective before/after amounts, rather the wallet balance.

parrot9design (approved):
LGTM 💪

PR #1160, opened 2023-02-02
Improve ICX queue UX

### LINKED ISSUE #1157: Update the ICX liquidity panel
The ICX pool has been causing some confusion because it really functions like a pool.

To address this, let's make some changes to the ICX-only view in the Supply Liquidity section:

ICX queue

- [ ] Change the header to: **ICX queue**

- [ ] Add a tooltip next to it with this message:

>The ICX queue facilitates "instant unstaking".
>
>You'll earn BALN while your ICX is in the queue. When you get to the front, your ICX will be exchanged for sICX, plus an extra 0.7%.
>
>You'll need to claim and unstake your sICX before you can supply it again.

(The 0.7% value should be pulled through via an on-chain API so we never have to manually update the number.)

- [ ] Display the BALN incentives and APY for the pool like the other incentivised pools (not sure when or why this was removed).

hetfly:
close #1157

### DIFF
--- a/src/app/components/Header/index.tsx
+++ b/src/app/components/Header/index.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
-import { CHAIN_INFO, SupportedChainId as NetworkId } from '@balancednetwork/balanced-js';
+import { BalancedJs, CHAIN_INFO, SupportedChainId as NetworkId } from '@balancednetwork/balanced-js';
 import { t, Trans } from '@lingui/macro';
+import BigNumber from 'bignumber.js';
 import { useIconReact } from 'packages/icon-react';
 import ClickAwayListener from 'react-click-away-listener';
 import { useMedia } from 'react-use';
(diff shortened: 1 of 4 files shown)


hetfly:
@parrot9design conversion fee is pulled from the Dex contract. Also added a notification to the wallet if the user has unstaked ICX

https://user-images.githubusercontent.com/13098459/216329430-1061b85c-a71f-458e-ace4-a647dcbd9a6d.mov

parrot9design:
Nicely done! Have unstaked some sICX to test the notification.

I did notice some weird behaviour with the ICX pool: The slider appears for one of my wallets, but when I tried two others that held more ICX, I had to fall back on the % options.

Screenshot 2023-02-03 at 8 21 36 AM

Screenshot 2023-02-03 at 8 21 54 AM

(The reason I tried with larger amounts of ICX was to see if the `Your daily rewards` data would stop showing `N/A`. It seems to stay static until after you've supplied ICX liquidity, whereas other pools update as you drag the slider.)

parrot9design (approved):
ICX is still unstaking so I can't test the notification, but otherwise this looks good to merge. Thanks @hetfly!

PR #1161, opened 2023-02-05
Feat/ remove activity history

Chunn:


parrot9design:
Hey @chunn or @hetfly: can you switch this to mainnet so we can test under normal conditions?

hetfly:
@parrot9design it's set to mainnet

parrot9design:
I haven't seen this behaviour before, but if you load the page while your wallet is set to testnet (even if still locked), it gets stuck in a refresh loop:

https://user-images.githubusercontent.com/63842577/221686732-75a9aca2-abfe-45e8-90d9-592ab3b3ab95.mov

It doesn't appear to do this in the app, so it must have been introduced here somehow?

---

The liquidity rewards tooltip is only half visible at many screen sizes:



The easiest solution would be to increase the padding at the bottom of the page. If we were to flip the tooltip, it would get in the way of the boost slider when in the adjust state.

BennyOptions:
@Chunn @RiseBlock just pinging this - Parrot9 found a few issues.

Chunn:
@parrot9design 
>I haven't seen this behaviour before, but if you load the page while your wallet is set to testnet (even if still locked), it gets stuck in a refresh loop:

Sorry. I've tried to reproduce this bug but it's not occurred. Could you provide reproduction steps for this bug?

https://user-images.githubusercontent.com/17588370/222946434-f6b04f5d-ec58-4e35-ba03-05d2bc65dea3.mov

parrot9design:
Interesting. If you have Hana set to `Mainnet` or the first `BTP testnet` on the list, there's no issue.

Screenshot 2023-03-06 at 4 36 05 PM

But if you set your wallet to any of the networks listed under `ICON`, then refresh the page, you should be able to reproduce the bug. In case it makes a difference, I've been testing in Brave.

Chunn:
@parrot9design I've tried to check based on your comment and also checked it in Brave. But I still can't reproduce this bug 😞. Please check the attached video:

https://user-images.githubusercontent.com/17588370/224547416-6fc4daf4-8124-4292-a7d2-fe82917c35ec.mov

<br/>

>The liquidity rewards tooltip is only half visible at many screen sizes:

I've fixed it. Please check again. Thank you!

parrot9design:
Strange. I still get that bug consistently, even when previewing the latest deployment link which is on the Berlin testnet (even if the wallet is also set to Berlin). Is the behaviour any different if you use the Vercel link instead of localhost?

In the console log, it keeps reporting ContractNotFound and 400 errors for https://berlin.net.solidwallet.io/api/v3, if that helps.

I checked the deployment link for another PR that was merged recently (https://github.com/balancednetwork/balanced-network-interface/pull/1172), and it seems to be an issue there, as well. I'm not seeing the same behaviour on app.balanced.network or test.balanced.network, though, so it should be safe to ignore this one for now. 

Are you able to set the branch to mainnet so I can review the latest changes?

hetfly (approved):

Chunn:
@parrot9design 
>Strange. I still get that bug consistently, even when previewing the latest deployment link which is on the Berlin testnet (even if the wallet is also set to Berlin). Is the behaviour any different if you use the Vercel link instead of localhost?

It seems to be the same. Please check the attached video that I checked on Vercel link:

https://user-images.githubusercontent.com/17588370/225039799-123d9e75-f05d-46fd-b098-208355db2956.mov

<br/>
>I checked the deployment link for another PR that was merged recently (https://github.com/balancednetwork/balanced-network-interface/pull/1172), and it seems to be an issue there, as well. I'm not seeing the same behaviour on app.balanced.network or test.balanced.network, though, so it should be safe to ignore this one for now.

OK. Thanks for sharing!

> Are you able to set the branch to mainnet so I can review the latest changes?

It seems it has been set to mainnet. Please check:
https://github.com/balancednetwork/balanced-network-interface/pull/1161#issuecomment-1446031624

parrot9design (approved):
Thanks for your work on this, @Chunn!

PR #1182, opened 2023-03-25
Fix: prevent swapping for same assets

Chunn:


parrot9design:
Hmm, I don't think removing the active token from the other asset dropdown is the way to go.

The default swap is BALN -> bnUSD, so if I want to swap bnUSD for another asset, I either have to reverse the swap or choose which asset to receive first. Neither of which are very intuitive.

We should display all assets in the dropdown, and if you choose an asset that's already active in the other field, the latter should revert to the previous token (essentially reversing the swap pair):

I don't think I'm explaining that well, so here's an example:

Field 1: BALN
Field 2: bnUSD

If you change Field 1 to bnUSD, Field 2 should default to BALN.

Chunn:
@parrot9design I got it. Thanks for your explanation

parrot9design:
Not sure why these PRs keep defaulting to the Berlin testnet. @chunn or @hetfly, can you please point this to mainnet to facilitate testing?

parrot9design (approved):

PR #1183, opened 2023-03-26
Fix/prevent opening Metamask when opening trade page

Chunn:


parrot9design (approved):
Much better, thanks for taking care of this. 🙏

PR #1192, opened 2023-04-10
Add arbitrary calls support

### LINKED ISSUE #1111: Add forum link to transaction data on vote page
This is now a required parameter in the transaction, please update the FE to account for this.

Screen Shot 2022-11-07 at 2 29 21 PM

hetfly:
close #1111


parrot9design:
For the forum link, let's change the link text from `See related forum post` to `View forum discussion`.

Screenshot 2023-04-18 at 1 02 41 PM

---

For new proposals, thinking we use these for the link text:

- `Add a contract call`
- `Copy data` / `Data copied`
- `Verify contract call(s)` (we can either keep the `(s)` in there permanently, or pluralise it once a second call is added)


We could potentially fit the last two on a single line, as below, but feel free to ignore me on this one if you don't think it's worth the effort.

`Add a contract call`
`Copy data` | `Verify contract call(s)`

---

For the `Verify contract call(s)` tooltip:
>Test your contract calls before you submit the proposal.
>
>**On Hana:**
>You'll see an error message before you sign the transaction. If it ends with **Reverted(20)**, the contract calls have been verified.
>
>**On Ledger:**
>Sign the transaction, then check the banner notification message for the result.

(I've assumed that the Ledger message will show something other than `Reverted(20)`? If not, let me know and I'll update the message.)


For the verification modal:

>**Verify contract calls?**
>
>Check that your contract calls are set up correctly before you submit the proposal.
>
>Go back | Verify

---

Let's also update the bBALN message to make it read like a normal sentence.
Screenshot 2023-04-18 at 1 23 28 PM Screenshot 2023-04-18 at 1 23 50 PM

>You must have at least {amount} bBALN to propose a change.

---

I don't want to test this properly as it's on mainnet now, but are you able to share the banner notification messages so I can make sure the wording is all good for those, as well?

hetfly:
@parrot9design 

Screenshot 2023-04-18 at 10 04 19
Screenshot 2023-04-18 at 10 05 19

When the verification fails, we can show only the generic error message:
Screenshot 2023-04-18 at 10 05 51

parrot9design:
Thank you! I think we can keep the banner notification content similar to the link text:

> Verifying contract calls...

> Contract calls verified.


And for the verification modal, let's give the text full opacity. The lower contrast is throwing me off a bit.
Screenshot 2023-04-19 at 5 35 43 PM

PR #1207, opened 2023-05-08
Add bribes UI

hetfly:


BennyOptions:
@hetfly is it possible to show an implied return for the votes instead of the nominal value? Or maybe both? As a user, the hard number is less important than the percentage return. This is what I was thinking:

Displayed return = totalBALNRewards (_currently displayed value_) / current bBALN votes allocated to it / number of days the bribe lasts * 365

@parrot9design also curious your thoughts on the design side.

Screenshot 2023-05-09 at 12 14 52 AM

parrot9design:
I hadn't considered showing an implied return, interesting idea! We could potentially add the APR next to the headings with the `label` style attached.

Are we likely to offer this much BALN for the bribes, though? Assuming we get 1,000,000 bBALN voting on the BALN/bnUSD pool, it would be less than a 2% return.

hetfly:
@parrot9design @BennyOptions  how about this:
```
APR = (nextWeekReward in USD * 52)  / (totalBBalnVoted * balnPrice) 
```

and the 1,000 bBALN yield for the next distribution:
```
yield = APR / 52 * 1000 * balnPrice
```

image

parrot9design:
I wonder if we should take an average of the 3 weeks shown, just in case they're not always consistent?

Thinking we could use this for the tooltip:

Screenshot 2023-05-12 at 10 36 34 AM

(I added a space between the tickers & the `/` to give it more room. Let's do that for all the pools in the liquidity incentives section & on the Stats page, as well – it'll make them consistent with how they're presented on the Trade page.)

parrot9design:
Assuming this is the rewards calculation we're using, I've added the details to the main tooltip for transparency:

Screenshot 2023-05-15 at 4 21 09 PM

>APR = (Average bribe (USD) x 52) ÷ (bBALN votes x BALN price)

Also just want to confirm that the bribe contract is set to release the bribes at that time (Thursday at 0:00 UTC)? I used it as a placeholder for the content, but didn't verify its accuracy.

---

For the pool headers, let's add a `:` after the pool and make the APR white to avoid blending in with the bribe info underneath (see screenshot above).

---

The `Claim bribe` modal looks a little empty, so let's enhance it a bit:
Screenshot 2023-05-15 at 4 46 09 PM

- Use `Claim bribe?` as the heading
- Make the bribe amount bold and 20px
- Reduce the modal width to 300px

---

Aside from that, this looks good to ship. @BennyOptions, are you wanting to push this out ASAP or wait until there's a proposal live for BALN bribes?

hetfly:
@parrot9design thanks, updated

PR #1225, opened 2023-07-24
Support links for trades and supplying

hetfly:


parrot9design:
For the pool/pair, let's separate the assets with an `_` instead of another `/` (`/trade/BALN_bnUSD` instead of `/trade/BALN/bnUSD`). Otherwise this looks good to go. 💪

PR #1229, opened 2023-08-01
Integrate Archway xCall

hetfly:


parrot9design:
Let's increase the padding for the ICON wallet hover state:
Screenshot 2023-11-06 at 4 20 17 PM

---

When at least one wallet has been connected, let's hide the disclaimer from the sign-in modal:
Screenshot 2023-10-27 at 2 13 18 PM

---

When signed out, let's update the message in the Activity section to `There are no pending cross-chain transactions.`

Screenshot 2023-11-06 at 4 30 06 PM

---

`Arch` should be `ARCH` here:
Screenshot 2023-11-06 at 4 46 18 PM

---

`Transfer Fee` should have a lowercase `f`. When there's no transfer fee, show `N/A`. 
Screenshot 2023-11-06 at 4 54 16 PM

Screenshot 2023-11-06 at 4 57 58 PM

---

If possible, let's try to keep the address to 2 lines here:

Screenshot 2023-11-06 at 4 46 54 PM

---

Not sure why `...` is shown here instead of `ICON`?
Screenshot 2023-11-06 at 4 51 34 PM

---

The IBC + xCall tooltip is a bit dense for my liking. Let's try this shortened version instead:

Screenshot 2023-11-06 at 5 54 57 PM


>**IBC** is the Cosmos interoperability protocol.
>
>**xCall** is a cross-chain messaging service that can interact with smart contracts on other blockchains.

parrot9design:
If you sign in on Archway, you currently can't see any of the Home page UI. Let's update it to look like this instead:
Screenshot 2023-11-24 at 1 17 03 PM

Collateral message:
>To deposit collateral, sign in on ICON.

Rewards message (display until users are eligible to earn rewards):
>To earn Balanced rewards, take out a loan, supply liquidity, or lock up BALN.

----

I can see in [this commit](https://github.com/balancednetwork/balanced-network-interface/pull/1229/commits/73d487596f2659bd2f774eb547e6e06d3a1b910f) that you updated the bridge tooltip content, but I'm still seeing the old message in the UI?

---

For sARCH transfers to Archway, let's try this for the modal design:
Screenshot 2023-11-24 at 3 12 24 PM

- Don't default to either option: deactivate the `Transfer` button until an option has been selected.
- Update sARCH to always use the blue logo instead of the orange version associated with non-staked ARCH.

parrot9design:
>If you sign in on Archway, you currently can't see any of the Home page UI. Let's update it to look like this instead

Thinking that we could also make this the default in the fully signed-out state, as well?

---

If you refresh the page, you're signed out of your Archway wallet. Can we stop this from happening?

---

It looks like we always need the recipient wallet to be connected before a transfer can go ahead? If so, let's use the same behaviour as the cross-chain swaps, with the `Connect wallet` prompt in the address field.

---

The initial `Approving...` and `Requesting...` banner messages trigger when you click the modal button. If possible, they should trigger when the user signs the transaction instead.

parrot9design:
The handling isn't quite right for the Transfer > Address field.
If you delete the prefilled address, the connected wallet address appears in green. You can enter an address to replace it, but if you submit it as-is (so the address is still green), it acts like there's no address to send it to.
Screenshot 2023-12-06 at 5 10 42 PM

If there's no reason to enter an address manually, let's make it behave like the address in the Swap section. (Although the fact that ICON > Archway swaps don't require confirmation on Archway now makes me wonder if people should be able to enter any recipient address, without connecting a wallet. 🤔)

---

Are we able to fix the chart to the bottom to prevent this white space?
Screenshot 2023-12-08 at 12 57 35 PM

---

The value isn't displaying correctly in the confirmation modal for cross-chain swaps from Archway (should be `1.6 ARCH`:
Screenshot 2023-12-08 at 2 12 40 PM

---

The confirmation modals often remain active even after the transaction is completed.
Screenshot 2023-12-08 at 2 18 06 PM

---

Will these numbers be hooked up to an API before launch? 
Screenshot 2023-12-08 at 2 20 43 PM

---

Can we always have the option to revoke a pending transaction? All my Archway transactions are failing right now so I can't complete this one. 
Screenshot 2023-12-08 at 2 23 53 PM

Possibly just like this?
slice1@2x

---

I tried a second transaction to Archway with the first still pending, and managed to break it. 😅  

The confirmation modal opened with the `Confirm` button active. So I clicked it.
Screenshot 2023-12-08 at 2 43 22 PM

The transaction failed, but I saw that I could still click the Transfer button, so I did.
Screenshot 2023-12-08 at 2 44 29 PM

I think I clicked Cancel from there and tried again, only to see behaviour like this:
Screenshot 2023-12-08 at 2 45 55 PM
Screenshot 2023-12-08 at 2 44 41 PM

I can reproduce it with all cross-chain swaps and transfers as long as there's a transaction pending. The only resolution was to quit my browser, which reset the transactions for some reason.

parrot9design:
Can we get this to say `0 bnUSD` instead of `undefined`?
Screenshot 2023-12-10 at 8 46 57 AM

---
Some of the pending notification messages end with a period instead of an ellipsis:
Screenshot 2023-12-10 at 8 40 54 AM
Screenshot 2023-11-29 at 4 56 09 PM

---

Change the `Increase allowance` button here to `Approve`:
Screenshot 2023-12-11 at 11 55 23 AM
Which is also a better fit for the banner message that follows it:
Screenshot 2023-12-11 at 11 55 33 AM

---

The wallet balance in the asset selector only reflects the ICON wallet. It should reflect the total across connected chains.

Screenshot 2023-12-11 at 12 00 11 PM

---
When signed out, I can still see the option to confirm transactions, although nothing happens if I click them.

Screenshot 2023-12-11 at 11 48 49 AM

---

I'm still having trouble confirming transactions on the Archway side, and it seems possible to keep stacking them up. 😅
Screenshot 2023-12-11 at 11 36 03 AM

Thinking that once the transaction is underway we change the `Cancel` button to `Revert`, so the transaction can be wiped clean to prevent this from happening:
Screenshot 2023-12-11 at 12 12 49 PM

---

This modal doesn't go away after you connect a wallet on mobile:

hetfly:
>Are we able to fix the chart to the bottom to prevent this white space?

It was quite buggy, right now I'd like to spend time on different things, but we can revisit this later. For now the chart is centered vertically.

>Will these numbers be hooked up to an API before launch?

Planning on it but there is a solid chance they won't.

>Can we always have the option to revert a pending transaction? 

Once xCall is in progress we have to wait for it to either succeed or fail. Can't turn it around in the middle.

>I tried a second transaction to Archway with the first still pending, and managed to break it.

This should be fixed now.

>When signed out, I can still see the option to confirm transactions, although nothing happens if I click them.

User is now prompted to log in in this case.

>This modal doesn't go away after you connect a wallet on mobile

This still remains.


Rest is fixed.

parrot9design:
>>Will these numbers be hooked up to an API before launch?
> 
> Planning on it but there is a solid chance they won't.

If we're not able to get the API hooked up in time, let's hide the top section altogether:

Screenshot 2023-12-13 at 5 59 49 PM

If we can get the API hooked up, let's fix these labels – should be `Cross-chain swaps` and `Cross-chain transfers` (some `s`'s are in the wrong place):
Screenshot 2023-12-13 at 10 38 10 AM

----

Every time I transfer assets, I find myself getting tripped up over the amount because we don't show the wallet balance upfront, and the asset selector shows a combined total. I keep entering more than I have and get confused as to why the Transfer button isn't active. 😅

What are your thoughts on adding the wallet balance above the asset, like we do in the Swap tab? (I know it's not the tidiest, but we'll clean this up considerably in our UI upgrade.)

bridge-shot 5 53 27 PM

----

Something funky is going on with the price impact. For 1 BALN > bnUSD, the price impact is 10%, but for 100 BALN, it's only 0.01%?
Screenshot 2023-12-13 at 10 35 36 AM Screenshot 2023-12-13 at 10 35 45 AM

----

Let's update the transaction failed content:

Screenshot 2023-12-12 at 9 18 17 AM

>Couldn't complete transaction. Revert requested.

Screenshot 2023-12-12 at 9 18 25 AM

>Revert transaction.
>[Revert]

I didn't quite catch the banner messages, so let me know if these don't match well enough:

>Reverting cross-chain {swap / transfer}...

>Cross-chain {swap / transfer} reverted.

FezBox:
SCR-20231214-lpxb

Hey everyone 

Long time user big time fan :-)

So after a bit of testing a little feedback for the cross chain swaps

The chain the asset is going to doest stand out enough, I found myself swapping to the wrong change a few times.
I feel that needs to be more prominent.

When you click to pick the chain it is very clear but if you didn't know you probably would overlook it.

parrot9design:
If you don't have an ICON wallet connected, let's hide the `New proposal` button on the Vote page:

Screenshot 2023-12-14 at 9 23 30 AM

---

In this state, I haven't connected an ICON wallet yet but it's not immediately clear why I can't transfer. Let's make the Transfer button active here so that we can prompt people to connect their wallet when they click it (like it is on the Swap tab).
Screenshot 2023-12-14 at 9 24 57 AM

---

`Arch` should be capitalised here:
Screenshot 2023-12-14 at 5 22 03 PM

---

Let's make the ARCH logo match the sARCH logo in terms of archway size:
Screenshot 2023-12-14 at 5 37 15 PM

---

I just realised that we don't allow people to send their Archway assets from the wallet. Is this possible, or is it too much to maintain for every additional blockchain?

Screenshot 2023-12-14 at 6 00 02 PM

parrot9design:
> SCR-20231214-lpxb
> 
> Hey everyone
> 
> Long time user big time fan :-)
> 
> So after a bit of testing a little feedback for the cross chain swaps
> 
> The chain the asset is going to doest stand out enough, I found myself swapping to the wrong change a few times. I feel that needs to be more prominent.
> 
> When you click to pick the chain it is very clear but if you didn't know you probably would overlook it.

Thanks @FezBox, appreciate the feedback!

I don't want to make any big changes to the UI, as it's quite a delicate balance as it is. However, after choosing an asset that's compatible with multiple blockchains, we could activate the blockchain selector to make people aware that they can also choose which chain to receive it on.

hetfly:
@parrot9design @FezBox 

>The chain the asset is going to doest stand out enough

We could add the blockchain logo next to it's name, similar to how it's in the opened dropdown, but that would make things pretty squeezed with longer chain names. So we could show chain logo only (keep dropdown with full names) or show logo and first two syllabels + '...' ?

parrot9design:
> @parrot9design @FezBox
> 
> > The chain the asset is going to doest stand out enough
> 
> We could add the blockchain logo next to it's name, similar to how it's in the opened dropdown, but that would make things pretty squeezed with longer chain names. So we could show chain logo only (keep dropdown with full names) or show logo and first two syllabels + '...' ?

Hmm, great suggestion, but the UI is already at its limit and additional logos will make it even more complex. We actually tried a variation like this early on in the design and quickly scrapped it. Let's stick with what I suggested in my previous comment unless that feels too clunky in practice.

hetfly:
>I just realised that we don't allow people to send their Archway assets from the wallet. Is this possible, or is it too much to maintain for every additional blockchain?

We can do it after the release.

FezBox:
> > SCR-20231214-lpxb
> > Hey everyone
> > Long time user big time fan :-)
> > So after a bit of testing a little feedback for the cross chain swaps
> > The chain the asset is going to doest stand out enough, I found myself swapping to the wrong change a few times. I feel that needs to be more prominent.
> > When you click to pick the chain it is very clear but if you didn't know you probably would overlook it.
> 
> Thanks @FezBox, appreciate the feedback!
> 
> I don't want to make any big changes to the UI, as it's quite a delicate balance as it is. However, after choosing an asset that's compatible with multiple blockchains, we could activate the blockchain selector to make people aware that they can also choose which chain to receive it on.

if it gets activated that should do the job. how would it work if doing a few swaps will the flow reset everytime? That would be my worry

hetfly:
@FezBox 

>how would it work if doing a few swaps will the flow reset everytime? That would be my worry

both the tokens and chains stay selected as they were

parrot9design:
Thanks @hetfly! It's close, but we can still tweak the behaviour a bit.

Let’s activate the blockchain selector after you choose a cross-chain asset from the dropdown, but at no other time.

Default to the native blockchain for each cross-chain asset, rather than always ICON.

When you reverse the swap, keep the previously selected chain and don't activate the blockchain selector.

parrot9design:
The liquidity slider isn't responding to text input. There's also an excessive number of decimals, maybe we should limit it to 6?

https://github.com/balancednetwork/balanced-network-interface/assets/63842577/4c7cd5ee-b306-4619-a4fd-88e3acbbcf98

---

Let's make cross-chain assets default to their native chain under the Supply tab, as well.

---

None of the BALN Incentive details are coming through for the sARCH pool yet, and it's not listed on the Trade page. Is that expected?

Screenshot 2023-12-17 at 2 56 50 PM

hetfly:
@parrot9design 

>Aside from Total daily rewards in the top section, none of the BALN Incentive details are coming through for the sARCH pool yet, and it's not listed under Incentivised pools. Is that expected?


It should fix itself when there is enough liquidity

PR #1309, opened 2024-04-01
remove legacy tokens - BUSD

swiftcc624:
@hetfly I removed BUSD, can you please let me know if there are any other legacy tokens?

here is the list of all tokens - https://docs.google.com/spreadsheets/d/1a6jl6RxB1sRYn9IwpyNjYi5YDCcHjJY3Ey6HNV98jlw/edit?usp=sharing

### DIFF
--- a/src/store/lists/communitylist.json
+++ b/src/store/lists/communitylist.json
@@ -43,14 +43,6 @@
       "logoURI": "https://raw.githubusercontent.com/balancednetwork/icons/master/tokens/crown.png"
     },
     {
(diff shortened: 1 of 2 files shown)


hetfly:
Thanks @swiftcc624 

default token list is fine as is now without the BUSD token, I'd say, but you can remove TAP, IAM, CLAW and CODE from community list https://github.com/balancednetwork/balanced-network-interface/blob/master/src/store/lists/communitylist.json

swiftcc624:
@hetfly I removed TAP, IAM, CLAW and CODE from community list

PR #1284, opened 2024-03-05
migrate to vite

octocat512:

PR #1316, opened 2024-04-12
Fix spinner issues

wilkyway-qx:
ref: https://github.com/balancednetwork/balanced-network-interface/pull/1284#issuecomment-2049226843


hetfly:
@wilkyway-qx @rise-techiast 
the last two points from the referenced comment are still not fixed

hetfly:
@wilkyway-qx @rise-techiast 

As this was part of the ledger refactor, also please implement keeping the address in local storage for an hour, same as is for browser wallet, storage key is `accountWithExpiry`. 

User should be signed in when the page is reloaded.

hetfly:
https://github.com/balancednetwork/balanced-network-interface/issues/1322

Issue #1201, opened 2023-04-19
Show LP fees APR in user liquidity details


hetfly:
Add Fees apr to the Current user supply section for  the users lp

hetfly:
@RiseBlock 
The fees APR is already available here:
<img width="1034" alt="Screenshot 2024-03-20 at 11 34 00" src="https://github.com/balancednetwork/balanced-network-interface/assets/13098459/15a5bd4e-ea7c-40eb-a489-b87c5886540e">

we want to add the fees APR to the users incentivised APR in the liquidity section, so that they can see the actual APR they are earning, instead of just BALN incentivised APR. Basically just add the pool fees APR to the users boosted incentivised APR and rename the column to APR here:
<img width="1055" alt="Screenshot 2024-03-20 at 11 37 49" src="https://github.com/balancednetwork/balanced-network-interface/assets/13098459/7d93f12b-4f4a-415e-bbf3-d34770eb0957">

we should also add the fees APR here:
<img width="866" alt="Screenshot 2024-03-20 at 11 37 12" src="https://github.com/balancednetwork/balanced-network-interface/assets/13098459/eb75e177-e0b8-45c2-84b9-33f8e2433eaa">

@parrot9design can you please second all that?

parrot9design:
The PR linked to this issue did not adequately address this issue. In future, do not merge any PRs until myself or @hetfly have reviewed them.

![Screenshot 2024-05-20 at 8 43 00 AM](https://github.com/balancednetwork/balanced-network-interface/assets/63842577/703773a9-3888-4f26-9aca-bf68a7fa778c)
- The BALN APR should not be a range here. It should be a specific number, based on the amount of bBALN I hold and the LP tokens I've staked.
- The APR column label needs a tooltip beside it like shown in the `Incentivised pools` list.
 - The tooltip message (in both places) needs to be adjusted to use `APR` instead of `APY`, and let's cut the line about impermanent loss.
![Screenshot 2024-05-20 at 8 50 01 AM](https://github.com/balancednetwork/balanced-network-interface/assets/63842577/b78fd291-e867-4de6-94fe-b6fafd0cdc94)

On mobile, I see an APY (should be APR) of 0.43%, which makes zero sense based on the numbers I see on desktop. As space is limited, it's fine to merge the fee APR and BALN APR into a single number.
[...]

PR #1335, opened 2024-05-26
fix: show lp fees APR in user liguity detail (#1201)

weeddotv:


hetfly:
Found two issues, otherwise looks good, thanks

- [ ] In the LP description section, there should still be APR range
Screenshot 2024-05-27 at 14 19 51
<hr />

- [ ] In the responsive layout, when the total APR is merged into a single number, you have to add the fees APR after multiplying the base by the boost, not before, otherwise wrong APR is shown.
Screenshot 2024-05-27 at 14 29 31

hetfly:
APR on phone is correct now, thanks!

- [ ] The range shows wrong numbers, same issue as previously. You have to add the fees APR **after** applying the boost.

sICX/bnUSD pair for example, the range should show `1.39% - 1.95%`

Screenshot 2024-05-31 at 11 31 06

<hr />

- [ ] Also, resolve conflicts please

PR #1355, opened 2024-06-19
feat: add base chain

0xmilktea:


FezBox:
@0xmilktea I tried to transfer eth from arbi to icon (or base)
clicked transfer its asks me to authorize after I click it then nothing happens 
image

parrot9design:
When I try to transfer USDC from ICON to Base, I get this message in Hana and can't move past it:

Screenshot 2024-06-25 at 4 31 07 PM

If I then try to test a USDC transfer to another chain (probably other transactions as well), Base messaging appeared in the modal during the transfer process:
Screenshot 2024-06-25 at 4 32 11 PM

And the transaction still shows as pending in the UI, even though the transaction is marked as approved on [xCallScan](https://xcallscan.xyz/messages/6324).
Screenshot 2024-06-25 at 4 40 50 PM

---

I suspect this behaviour is due to an issue in how transactions are handled when the wallet flags an issue. It's definitely a problem on mainnet as well so it doesn't _need_ to be resolved in this PR, but it is reproducible, so it will help us address the residual UI issues around cross-chain transaction activity.

parrot9design:
I'm unable to transfer USDC to Base on any chain. From ICON, the wallet stops it. From Arbitrum & Avalanche, the transaction rolls back halfway through. @FezBox this looks to be a smart contract issue.
   Screenshot 2024-06-25 at 5 29 09 PM

- [ ] The `REMOVE` link should also use regular casing (`Remove`).

parrot9design:
Let's update the Base logo to use this variation instead:
base

parrot9design:
If you choose to swap 100% of your ETH from Base, the frontend doesn't account for the gas fee so the Swap button in the confirmation modal won't trigger the wallet.

The gas fee is also ignored on the Bridge tab, where you can choose to send 100% of your ETH from Base and Arbitrum, and AVAX from Avalanche.

parrot9design:
- [ ] Let's remove the link from the message when the amount is 0 (because there's no reason to click it):

Screenshot 2024-06-27 at 4 35 12 PM

---

- [ ] Increase the message width so it remains on a single line:

Screenshot 2024-06-27 at 4 32 59 PM

parrot9design:
When you try to transfer from an EVM chain that isn't the active one, typing in an amount greater than the wallet balance won't trigger the `Insufficient {ticker}` message. You can open the confirmation modal and switch networks, but then nothing happens when you click Transfer.

PR #1347, opened 2024-06-15
Add native cross chain loans UI

hetfly:

PR #1357, opened 2024-06-20
Update wallet to show all EVM chains

0xmilktea:


parrot9design:
We'll resolve this from a design perspective in #1347. It's only a couple weeks away, so I don't think a stopgap design is required in the interim.

PR #1362, opened 2024-07-09
fix: btc

0xmilktea:


0xmilktea:
icx balance is not updating. check it

FezBox:
Please update so that there is only one BTC for all chains 
when the user goes to swap for BTC they pick the chain and get that version of BTC on that chain 
(exactly like how eth works)

parrot9design:
Looking good! 

Let's just replace '**BTCB(Old)**' with '**BTCB** (old)' (note the use of spacing, lowercase, and unbolded text), and omit `(old)` from the wallet balance above:
   Screenshot 2024-07-15 at 1 57 50 PM

PR #1384, opened 2024-07-30
feat: add usdt on arb and avax

0xmilktea:


parrot9design:
- [ ] Remove `Only` from this message when the amount is 0:

Screenshot 2024-07-30 at 3 49 13 PM

>0 {ticker} is available on {blockchain}.

I tested swapping USDT for bnUSD, AVAX, and BTC, and transferred USDT round a few chains, all went smoothly. After the wording issue is fixed, this should be ready to merge.

parrot9design:
Looks like that didn't fix the issue:
Screenshot 2024-07-31 at 2 16 31 PM

I think it's not accounting for decimals, as the message disappears if I enter `0.01`.

PR #1390, opened 2024-08-02
Refactor XLoans

hetfly:


hetfly:
@0xmilktea regarding the two `ChainList` components - they use different hooks, different data, have different column headers, different logic on item click, one lets you connect to a chain, the other does not. I'd prefer to keep them separated.

The rest is refactored, thanks for your feedback 👍

PR #1392, opened 2024-08-02
use nodePolyfills from a differnt package, use @ for absolute path

swiftcc624:


0xmilktea:
looks good @swiftcc624

PR #1415, opened 2024-08-21
Fix manual address input

### LINKED ISSUE #1416: Fix manual address input for Swap and Bridge
Somewhere during recent refactoring this stopped working
image

hetfly:
close #1416


apps/web/src/hooks/useManualAddresses.tsx
```diff
@@ -0,0 +1,15 @@
+import { XChainId } from '@/types';
+import { useCallback, useState } from 'react';
+
+export default function useManualAddresses(): {
+  manualAddresses: { [xChainId: string]: string | undefined };
+  setManualAddress: (xChainId: XChainId, address: string | undefined) => void;
```
0xmilktea, line 6: is the hook still useful?

parrot9design (approved):
Much better. Thanks for taking care of this.

parrot9design:
Actually, I did find an issue when signed in with an ICON, Havah, and/or Archway wallet: If you enter a custom address for, say, BNB Chain, it keeps the custom address for all other EVM chains (expected), but also for Injective. It doesn't register that it's an invalid address for Injective until I click to change it.

hetfly:
Fixed the injective issue, but it was part of the (IMO fault) logic that kept manual chain address across compatible chains. I think it might be better to keep the manually added address only per currently active chain, though, meaning that if I input BNB address, then change to arbitrum, it defaults to logged in wallet, rather than the BNB edited address.

parrot9design:
Thanks @hetfly, I'm happy with that handling.

One thing I found myself wanting during testing is the ability to hit Enter rather than click Use for the custom address. Is it possible to add that? Otherwise this is good to merge.

PR #1436, opened 2024-09-02
Update stats page

hetfly:
Issue is recorded in the old stats page repo
https://github.com/balancednetwork/balanced-network-info/issues/214

### DIFF
--- a/apps/stats/src/pages/PerformanceDetails/index.tsx
+++ b/apps/stats/src/pages/PerformanceDetails/index.tsx
@@ -9,11 +9,11 @@ import Header from '@/components/Header';
 import { Container, Divider } from '@/pages/StatsPage';
 import { Typography } from '@/theme';
 
+import { useLocation } from 'react-router-dom';
 // import BBALNSection from './sections/BBALNSection';
 import EarningsSection from './sections/EarningSection';
 import HoldingsSection from './sections/HoldingsSection';
 import StabilityFundSection from './sections/StabilityFundSection';
-import { useLocation } from 'react-router-dom';
 
 export const GridItem = styled.div`
   text-align: right;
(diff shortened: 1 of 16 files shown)


parrot9design:
For the network-owned liquidity chart, let's update these colours:
AVAX: `#FF394A`
sICX: `#9D4DF1`
SUI: `#4DA2FF`
INJ: `#0082FA`

Some of the colours clash a bit with the current organisation, though. Can we make all the charts start at 0° and move clockwise from largest to smallest amount? Might also help (or make it much harder 😅) to prevent the labels from overlapping: 
Screenshot 2024-09-11 at 10 30 21 AM

---

This tooltip has the fee hardcoded, as it should be 0.3% instead of 0.5%.
Screenshot 2024-09-11 at 10 47 13 AM


---

What's the threshold for the Reserve Fund chart? There are 7 tokens in it currently, but we only show the top 3.

If we want to use the asset colours instead of the Balanced colour palette for this chart, we'll need to use the sICX purple and update the DAO Fund chart colours for consistency (bnUSD: `#32C9A4`; BALN: `#2CA9B7`). Not sure I'm a big fan of the predominantly purple look, although it does break it up a bit. What do you think?

**Current** | **Asset colours** | **Balanced colours**
----------- | ------------------ | -------------------
Screenshot 2024-09-11 at 10 44 45 AM | Screenshot 2024-09-11 at 10 56 19 AM | Screenshot 2024-09-11 at 10 57 37 AM

hetfly:
>Can we make all the charts start at 0° and move clockwise from largest to smallest amount?

Good idea, at the moment it helps with the labels for smaller amounts being on the side. But in the future we might be forced to either position the items manually or group the smaller holdings to a single `others` item with some explanation.

>Not sure I'm a big fan of the predominantly purple look, although it does break it up a bit. What do you think?

I've updated the colours for both Reserve and DAO fund chart. I'm kinda liking using an asset specific colour, but I definitely see what you're saying 😅 Let's keep it this way for a while and see how it feels? We will have to do a similar thing sooner or later with all the new integrations coming anyway.

parrot9design (approved):
I think it needs to be updated from main to include the Token section update that's live on the Stats page currently? Otherwise this looks good to go, thanks @hetfly.

PR #1438, opened 2024-09-03
Remove standard tx popups for xcalls

### LINKED ISSUE #1358: Improve or remove the banner messages from cross-chain transactions
The banner messages for cross-chain transactions report that a transaction is successful when only the first half is complete. It's somewhat confusing at best and alarming at worst.

We should be able to detect when the transaction is fully complete because the UI has to update accordingly. Let's use the same trigger to serve the success messages at the appropriate time,. 

If it's not as easy to implement this as I've assumed, we should remove the messages from all collateral/loan/swap/bridge transactions.

We can leave the messages in place for ICON <> ICON transactions.

### LINKED ISSUE #1422: Simplify the modal messages when a cross-chain transaction is in progress
The cross-chain modals are a little too verbose about what's going on while a transaction is processing, e.g. exposing the ICON blockchain when it's not one of the chains people chose to interact with.

Instead of using this sequence for the messaging:

1. Awaiting confirmation on {blockchain 1}…
2. Finalising transaction on ICON…
3. Awaiting confirmation on ICON…
4. Finalising transaction on {blockchain 2}…

Let's just do this:

1. Confirming transaction on {blockchain 1}…
2. Finalising transaction on {blockchain 2}…

hetfly:
close #1358
close #1422


hetfly:
@parrot9design regarding the modal message simplification, it's a bit tricky. The issue you're describing with showing notifications related to ICON chain while transacting on two different chains, consists of multiple xCalls. One xCall always consists of 'confirm on origin chain' and 'finalise on destination chain'. That's where the ICON notifications come from, it always gets finalised on ICON, after that some transactions invoke another xCall (nonICON to nonICON swaps, collateral withdraw to another chain...)

So I landed on this:
For the first xCall, it always says only 'confirming on {origin chain}',
for the additionally invoked xCall, it says finalising on {destination chain}

parrot9design (approved):
Looks good, thanks @hetfly!

PR #1444, opened 2024-09-07
Update the wallet dropdown and asset lists

### LINKED ISSUE #1403: Update the wallet dropdown and asset lists
- [x] Add a chevron next to ICON-based assets in the wallet dropdown to indicate that you can interact with them
icon-assets

<br>

- [x] Add a footer row to the wallet dropdown to show the total holdings across connected wallets.

<br>

- [ ] The background that appears when you hover over the scrollbar in the Sign In / Manage Wallets modal is much nicer on the eye on than the harsh white background that appears everywhere else. Can we update all scrollable lists to use this style?

Ideal:

To fix:
 

<br>

- [x] Add decimal support for HVH in the wallet
Screenshot 2024-08-19 at 4 43 13 PM

hetfly:
close #1403


hetfly:
@parrot9design Ad the scrollbar styles - I think that all the scrollbars are behaving the same. The nicer style is active when you stay focused in the dropdown. When you move the focus out of it, it activates the ugly styles. Is this the case for you too?

VIdeo here https://streamable.com/skfgqr

parrot9design:
Your video is no longer available, but here's the difference I see with the scrollbar between the sign-in modal and the others:
https://github.com/user-attachments/assets/644ce5e9-2da1-45fe-8e8a-162db2485e1a

Note that I have my OS set to only display scroll bars when scrolling. When set to always show the scroll bar, it's the same white style everywhere. Is it quite difficult to customise the style, even moderately? It looks awful whenever someone posts a screenshot with the scroll bar in view, would be nice to harmonise it with the UI. 😅

---

- [ ] Can we keep the wallet modal open in the background when you click an ICON-based asset?

---

- [ ] Let's update the wallet search placeholder text to `Search assets and blockchains`

Screenshot 2024-09-11 at 2 46 07 PM

hetfly:
>Is it quite difficult to customise the style, even moderately?

It was my PTSD from the days when handling scrollbars basically meant using javascript, which usually affected the scrolling experience itself (hardly in a good way) and looked differently in each browser anyway 😅 But now it looks like the support in modern browsers is quite decent, so it looks like we might have a good solution. Let me know how this feels @parrot9design  👍

parrot9design:
Legend! That looks so much better, appreciate you looking into this again. Let's update the scrollbar colour to `rgb(20, 74, 104)`, then it should be good to ship. 💪

PR #1451, opened 2024-09-17
Improve asset selector when signed off

hetfly:


parrot9design:
Let's use the signed-out state for the `All assets` view also. It won't include your wallet breakdown, but it will make the list of supported assets easier to browse.

parrot9design:
Can you replace the `Wallet` column with `Price` in the `All assets` view (so it's exactly the same info you see in the signed out state)?

parrot9design (approved):

PR #1447, opened 2024-09-09
Cross chain loan enhancements

hetfly:
close #1401

PR #1481, opened 2024-10-07
Add stellar

hetfly:


0xmilktea (approved):

parrot9design (approved):
Looks good, thanks @hetfly!

I'll keep testing, but at this stage we just need to wait for the audit and XLM liquidity before we merge.

FezBox:
ok lets spend today seeing if we can optimise these @hetfly

0xmilktea:
image
@hetfly

parrot9design:
If you try to start a swap transaction from an inactive Stellar wallet, you can still click through to the confirmation modal:

Screenshot 2024-11-28 at 8 02 49 AM

We can handle this the same way as the inverse, by deactivating the Swap button and adding the error message underneath.

parrot9design:
When in the signed out state, the button should keep the `Swap` label after you enter an amount. It now says `Insufficient {asset}` instead

Live app | This PR
-------- | --------
Screenshot 2024-11-28 at 6 03 30 PM | Screenshot 2024-11-28 at 6 03 17 PM


If that's too much of a PITA to fix before we launch, let's tackle it in https://github.com/balancednetwork/balanced-network-interface/pull/1447 and get that merged as soon as we can afterwards.

PR #1489, opened 2024-10-10
feat: add injective USDC

0xmilktea:


parrot9design:
- [ ] When I click the max amount available on Injective, it fills the value but it doesn't allow me to proceed with the swap until I enter a smaller amount.

https://github.com/user-attachments/assets/dde61cf2-1a17-442e-97e5-a2758e6bef61

- [ ] Let's remove the `is` from that string: `Only {amount} {ticker} available on {blockchain}.`


Once those are fixed, this should be good to merge.

parrot9design:
Looks good to me, thanks @0xmilktea.

PR #1514, opened 2024-10-24
Show all pools instead of just the incentivised ones

### LINKED ISSUE #1512: Update the Liquidity Pools section to show all pools, rather than just BALN-incentivised ones
We'll also need to change the section toggle from `Incentivised pools` to `All pools` when signed in.

hetfly:
close #1512

### DIFF
--- a/apps/web/src/app/pages/trade/supply/_components/AllPoolsPanel.tsx
+++ b/apps/web/src/app/pages/trade/supply/_components/AllPoolsPanel.tsx
@@ -10,7 +10,7 @@ import { MouseoverTooltip } from '@/app/components/Tooltip';
 import { Typography } from '@/app/theme';
 import QuestionIcon from '@/assets/icons/question.svg';
 import useSort from '@/hooks/useSort';
-import { MIN_LIQUIDITY_TO_INCLUDE, PairData, useAllPairsById } from '@/queries/backendv2';
+import { MIN_LIQUIDITY_TO_INCLUDE, PairData, useAllPairsById, useNOLPools } from '@/queries/backendv2';
 import { useDerivedMintInfo, useMintActionHandlers } from '@/store/mint/hooks';
 import { Field } from '@/store/mint/reducer';
 import { getFormattedNumber } from '@/utils/formatter';
@@ -20,6 +20,7 @@ import { HeaderText } from '@/app/components/SearchModal/styleds';
 import Skeleton from '@/app/components/Skeleton';
 import { MAX_BOOST } from '@/app/components/home/BBaln/utils';
 import { PairInfo } from '@/types';
+import { xChainMap } from '@/xwagmi/constants/xChains';
 
 const COMPACT_ITEM_COUNT = 8;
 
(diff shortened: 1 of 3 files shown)


parrot9design:
I remember now why we didn't do this earlier. Do we really want all of these pools to appear on the page? (cc @FezBox)
Screenshot 2024-10-25 at 12 03 39 PM

Also, how is this pool getting BALN incentives? It's not on the list of pools on the vote page.
Screenshot 2024-10-25 at 11 43 42 AM


In any case, the list is long enough now that we should add a search box to the top right corner.

FezBox:
All key /bnusd pairs should appear there.


Or by default all nol and new chain native asset pairs should be incentivised. Or all those pairs should be there.

We should not have to go to another page to figure out pools when on the supply section of the page

parrot9design:
Ok, so let's go with incentivised pools + NOL and native asset pairs, as that gives us the best coverage to relevance ratio.

That basically means adding INJ/bnUSD, SUI/bnUSD, sARCH/bnUSD, and BNB/bnUSD to the list current in the app, with allowance for any new pools to appear automatically.

hetfly:
@parrot9design updated the list

also @FezBox , when you type in `nol` or `native` in the search box  it shows only NOL pools or pools with native tokens respectively.

parrot9design:
Nice, thanks @hetfly! Wondering if we change `All pools` to `Core pools`, since it's not really all of them now. Otherwise this looks good to go.

I assume the SOL/bnUSD pool will appear under the `native` search term once you can trade it on Solana?

hetfly:
@parrot9design yeah, SOL pool will be there once the SOL PR gets merged. 

I think keeping the tab called `All pools` is fine. In the end it is all the significant pools, perhaps we can include some liquidity filter for the rest of them, maybe > $5k, that would show CROWN, GBET and FRMD, which are still relatively alive projects.

parrot9design:
>I think keeping the tab called All pools is fine. In the end it is all the significant pools, perhaps we can include some liquidity filter for the rest of them, maybe > $5k, that would show CROWN, GBET and FRMD, which are still relatively alive projects.

My main concern is that we'd be acting like the list includes all pools when it doesn't. Even with a filter of >$5k, it would allow pools like IUSDC / bnUSD, sICX / BTCB, and OMM / IUSDC to appear, when they really shouldn't. At least, not unless you have the community token list active.

OMM / sICX probably should be included, though.

FezBox:
Search field makes a huge difference. thank you

hetfly:
>My main concern is that we'd be acting like the list includes all pools when it doesn't. Even with a filter of >$5k, it would allow pools like IUSDC / bnUSD, sICX / BTCB, and OMM / IUSDC to appear, when they really shouldn't. At least, not unless you have the community token list active.

@parrot9design we have a 'token blacklist' setup, tokens on the list won't have their pools shown. IUSDC, BTCB and USDS tokens are there. `Core pools` provides clearer messaging, though, so I don't mind using this label.

parrot9design:
Let's go with `Core pools` for now, we can always adjust later if needed.

Also, currently getting a white screen on the Trade page from the current preview link.

PR #1519, opened 2024-10-25
add USDC on Sui

0xmilktea:


parrot9design:
Tried to swap and transfer Sui USDC, but all transactions failed and the USDC has not yet reappeared in my wallet.

https://xcallscan.xyz/messages/18252
https://xcallscan.xyz/messages/18253
https://xcallscan.xyz/messages/18254

Issue #678, opened 2021-09-15
Charts: Build implied charts even without a pool


BennyOptions:
With transaction routing launching, we need to consider how to handle chart displays since we can't display every single possibility, especially as we get more tokens

parrot9design:
Long term we could include a dropdown selector on the chart itself, so people can choose the pair they want to see. But that will take a bit of work on the logic side to get working correctly, so it’s too ambitious as a solution just yet.

For now, the simplest approach would be to handle it similar to when we launched, without any price chart functionality.

- If there’s a pool for the assets, load the chart.
- If there isn’t a pool for the assets, display a message where the chart should be:

![DBACC215-BD67-4B95-B6E7-1D08F3C7F40F](https://user-images.githubusercontent.com/63842577/133532147-c863dd0f-f280-4534-929e-c19e84292a04.jpeg)

> No price chart available for this pair.

BennyOptions:
will be fixed with code refactoring

BennyOptions:
Could maybe emit an event log for swapping and build a database of cross-pool swaps for charts of trades without a liquidity pool

PR #1522, opened 2024-11-01
Trade page enhancements

### LINKED ISSUE #1402: Trade page enhancements
- [x] Make the bridge activity section scrollable so it doesn't increase the height of the entire panel when there are 3+ transactions pending.
Screenshot 2024-08-19 at 3 21 02 PM

<br>

- [x] Default to 1D on the price charts.
- [x] Update the TradingView charts to use these time frames: 15m, 1H, 4H, 1D, 1W, 1M. Make 1D the default.
- [x] Remove the TradingView chart option for assets on the community token list, e.g. GBET, which don't show any data
- [x] Fix the price decimals on some of the pairs:
Screenshot 2024-09-17 at 10 17 05 AM

<br>
<br>

- [x] Prevent the Swap For asset selector from repositioning itself as you type.
https://github.com/user-attachments/assets/3ba82962-df8c-4eb1-8dcf-1e6e81913c49

<br>

- [x] Fix the overflow issue that appears when the Supply tab is active on mobile

<br>
<br>

- [x] Update the fee message in the confirmation modals:

balanced-network-interface-avvkaja3o-balanced-dao vercel app_ balanced-network-interface-avvkaja3o-balanced-dao vercel app_ (1)
I had mixed results last time I tested this, but if the swap fee is included in the amount (desired behaviour) rather than additional, use `Swap fee (included)`. 

balanced-network-interface-avvkaja3o-balanced-dao vercel app_trade_bridge

<br>

- [x] Update the swap metadata to clarify the fees (`Swap fee` and `Transfer fee`):
Screenshot 2024-08-19 at 4 05 44 PM

<br>

- [x] Update the BTC decimals so we can display a non-zero exchange rate for bnUSD > BTC
Screenshot 2024-08-19 at 4 21 40 PM

<br>

- [x] Fix the datapoints in the Activity section, which rarely seem to load in any data:

Screenshot 2024-08-19 at 4 39 10 PM

<br>

- [x] #1512
Screenshot 2024-08-28 at 3 45 22 PM

<br>

- [x] If you're not connected to the chain you try to transact from but the receiving address is known, the Swap & Transfer buttons should trigger the chain-specific connect wallet prompt. E.g. in these situations, only the transacting wallet is unknown, yet it triggers the full sign-in modal.
Screenshot 2024-08-28 at 5 16 56 PM Screenshot 2024-08-28 at 5 16 47 PM

<br>

- [x] Limit the number of decimals included when you click the 25/50/75/100 options.
Screenshot 2024-10-07 at 3 46 09 PM

- [x] Fix the decimal handling for large asset amounts in the asset selectors. 5 is excessive here, especially for the sICX entry.
Screenshot 2024-10-07 at 3 48 27 PM

hetfly:
close #1402

### DIFF
--- a/apps/web/src/app/components/CollateralTypeSwitcher/CollateralTypeList.tsx
+++ b/apps/web/src/app/components/CollateralTypeSwitcher/CollateralTypeList.tsx
@@ -5,6 +5,7 @@ import { useMedia } from 'react-use';
 import { Typography } from '@/app/theme';
 import { useAllCollateralData, useCollateralActionHandlers, useUserPositionsData } from '@/store/collateral/hooks';
 import { useLoanActionHandlers } from '@/store/loan/hooks';
+import { getSpokeVersions } from '@/utils/xTokens';
 import { ICON_XCALL_NETWORK_ID } from '@/xwagmi/constants';
 import { xChainMap } from '@/xwagmi/constants/xChains';
 import { XChainId } from '@/xwagmi/types';
@@ -54,6 +55,7 @@ const CollateralTypeList = ({
       xPosition =>
         xPosition.baseToken.symbol.toLowerCase().includes(query.toLowerCase()) ||
         xPosition.baseToken.name?.toLowerCase().includes(query.toLowerCase()) ||
+        getSpokeVersions(xPosition.baseToken.symbol).some(v => v.toLowerCase().includes(query.toLowerCase())) ||
         Object.keys(xPosition.positions).some(x => xChainMap[x].name.toLowerCase().includes(query.toLowerCase())),
     );
   }, [positions, query]);
--- a/apps/web/src/app/components/CurrencyInputPanel/index.tsx
+++ b/apps/web/src/app/components/CurrencyInputPanel/index.tsx
@@ -9,8 +9,10 @@ import CurrencyLogo from '@/app/components/CurrencyLogo';
 import { SelectorPopover } from '@/app/components/Popover';
 import DropDown from '@/assets/icons/arrow-down.svg';
 import useWidth from '@/hooks/useWidth';
+import { useRatesWithOracle } from '@/queries/reward';
 import { COMMON_PERCENTS } from '@/store/swap/reducer';
 import { escapeRegExp } from '@/utils';
+import { formatBalance } from '@/utils/formatter';
 import { DEFAULT_TOKEN_CHAIN } from '@/xwagmi/constants/xTokens';
 import { XChainId } from '@/xwagmi/types';
 import { getSupportedXChainForToken } from '@/xwagmi/xcall/utils';
@@ -147,6 +149,13 @@ export default function CurrencyInputPanel({
 
   const [ref, width] = useWidth();
 
+  const prices = useRatesWithOracle();
+  const price = useMemo(() => {
+    if (prices && currency?.symbol) {
+      return prices[currency.symbol];
+    }
+  }, [prices, currency]);
+
   const handlePercentSelect = (instant: number) => (e: React.MouseEvent) => {
     onPercentSelect && onPercentSelect(instant);
   };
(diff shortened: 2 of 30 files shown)


hetfly:
@parrot9design 

>Update the TradingView charts to use these time frames: 15m, 1H, 4H, 1D, 1W, 1M.

I did add the 1M option, but the candles are broken. We don't have that granularity available from BE and TradingView recalculation is having some troubles. I'd remove the 1M interval from TV options.

<hr />

>Fix the overflow issue that appears when the Supply tab is active on mobile

Some black magic was happening on mobile with the APR tooltip in the pools table header. Had to hide it on mobile, haven't found a workaround, unfortunately. 

<hr />

>Limit the number of decimals included when you click the 25/50/75/100 options.

Done, but it only is visual, though. In the confirmation modal, full amount is displayed. I think it's better to keep the precision at least there, since otherwise we wouldn't be showing the exact amount being sent to the contract anywhere else.

<hr />

Apart from the items listed in the connected issue, I've also implemented @dbrehmer suggestion to show oracle price for particular tokens (currently USDC only) in the token list on the swap page, hid the charts for stable/bnUSD (it shows 'false' price movements, stable coin swaps are handled through the stability fund) and hid the untradeable tokens from swap selectors (currently tBTC, weETH and wstETH, in the future usually tokens used for minting with no liquidity), as specified here https://github.com/balancednetwork/balanced-network-interface/issues/1523

parrot9design:
>I did add the 1M option, but the candles are broken. We don't have that granularity available from BE and TradingView recalculation is having some troubles. I'd remove the 1M interval from TV options.

Yeah, let's remove the 1M option, there's not enough data to make it valuable yet.

---

- [ ] Is it possible to get the charts to display more data when you scroll backwards?

https://github.com/user-attachments/assets/e687523c-f23e-414c-b57a-f9b4823c0e9f

---

- [ ] Should it be trying (and failing) to pull through the exchange rate for pairs that don't have a price chart? Or are we preparing to finally implement #678? 😉 


Screenshot 2024-11-06 at 3 43 16 PM

---

- [ ] Thoughts on hiding the decimals for values over 100? Fractions are insignificant once you hold 100+ of them.
Screenshot 2024-11-06 at 3 47 07 PM


---

- [ ] If you try to swap on the same chain, the sign-in modal appears instead of the confirmation modal, even when signed in.

https://github.com/user-attachments/assets/dbee9526-e423-4df2-bfe0-8defc7aa94e3


---

- [ ] Can we add cbBTC, wBTC, BTCB (the new version) to the list of search terms (trade page and collateral)?

hetfly:
@parrot9design 

>Is it possible to get the charts to display more data when you scroll backwards?

There is no support built for this currently, at least it's included in the trading view option, though. 

<hr />

>are we preparing to finally implement implied charts?

Next proposal item, I think.

<hr />

> Thoughts on hiding the decimals for values over 100? Fractions are insignificant once you hold 100+ of them.

Sounds good, let's try it out

<hr />

>Can we add cbBTC, wBTC, BTCB (the new version) to the list of search terms (trade page and collateral)?

Only in a way that would show the same results as when you search for BTC - in other words, the other BTC variant's would not be hidden. Otherwise it would be quite a lot of code and effort for what it's supposed to do.

parrot9design:
The decimals are much cleaner, though can we also remove them when the value is above $100?
Screenshot 2024-11-11 at 12 29 20 PM Screenshot 2024-11-11 at 12 29 42 PM


<br>

>    >Can we add cbBTC, wBTC, BTCB (the new version) to the list of search terms (trade page and collateral)?
> 
> Only in a way that would show the same results as when you search for BTC - in other words, the other BTC variant's would not be hidden. Otherwise it would be quite a lot of code and effort for what it's supposed to do.

I'm happy to try that for now, at least until we can get clearer handling like this in place (discussed in Discord): 
image



Otherwise, I think it's good to go. Thanks for your work on this.

PR #1524, opened 2024-11-06
Stats: Remove METX and FIN

hetfly:


hetfly:
@parrot9design Tooltips on mobile are hidden and the horizontal overflow issue seems to be gone. Are there any tooltips you would be really missing and want me to look for some workaround?

parrot9design:
Excellent, thank you!

We should try to include these tooltips if possible:
Screenshot 2024-11-07 at 6 39 24 AM

Maybe also the tooltips next to the `ICON enshrinement` and `Withdrawal limit` sections.

PR #1536, opened 2024-11-20
Implement wICX

### LINKED ISSUE #1534: wICX implementation

hetfly:
close #1534


hetfly:
@FezBox managed to temporarily put in a check for a better trade using the queue while there is still liquidity. We will have to do a [clean up ](https://github.com/balancednetwork/balanced-network-interface/issues/1538)after the queue is emptied and deprecated anyway, there is lot's of conditional code using the queue. 

Can we get some liquidity in the new pool so that we can properly test the routing and trade picking? (queue vs wICX)

@parrot9design only when supplying and trading into ICX, wICX contract is used, but it's all abstracted from the user and `ICX` symbol is displayed instead of `wICX`. While we have the queue working in parallel temporarily, do we want to differentiate the wICX contract somehow? At least in `Your pools` section, we could show `wICX/sICX` as the pool name? If the user had both, the queue and wICX/sICX in that section, it could look a bit confusing.

parrot9design:
Let's use ICX / sICX to identify the new pool so we can omit all mention of `wICX` from the UI.

We can call the old one the ICX queue:
Screenshot 2024-11-22 at 1 05 10 PM

parrot9design:
There are still a few mentions of wICX in the UI:
Screenshot 2024-11-25 at 10 29 22 AM Screenshot 2024-11-25 at 10 29 32 AM
Screenshot 2024-11-25 at 10 32 46 AM


The asset selectors treat ICX as though it's wICX, as it should be above sICX when the list is ordered alphabetically.
Screenshot 2024-11-25 at 10 28 39 AM

When you choose ICX, can we make the second asset default to sICX instead of bnUSD?

Also, shouldn't we be able to stake LP tokens for the new pool?
Screenshot 2024-11-25 at 10 36 59 AM

hetfly:
Left to do:

1. test swaps to ICX with liquidity in the new pool
2. decide on whether we stake ICX when trading if the wICX/sICX pool has worse rate
3. decide on having unstaking ICX as an option while withdrawing ICX collateral

parrot9design:
- [ ] Remove the `w` from `wICX` here:
Screenshot 2024-12-17 at 10 14 41 AM


Otherwise, we just need to resolve the conflicts with the master branch.

PR #1544, opened 2024-11-27
Add POL rewards support

### LINKED ISSUE #1367: Proof of liquidity Rewards rework
Rewards on Balanced will not also accept external reward tokens.
Initially it will be sICX from emissions.
The update did not seem to introduce any breaking changes that i could find from testnet UI but should be double checked.
Get Baln holding will be replaced with getRewards, which will return a map of  token Address => amount
    @External(readonly = true)
    BigInteger getBalnHolding(String _holder);

   @External(readonly = true)
    Map<String, BigInteger> getRewards(String _holder);

Removed method that could be checked for
    BigInteger getAPY(String _name);
    Map<String, BigInteger> recipientAt(BigInteger _day);
    List<String> getRecipients();
    Map<String, Object> distStatus();
    Map<String, BigInteger> getRecipientsSplit();

New APY calculations: to be discussed when we get to it

hetfly:
close #1367


hetfly:
@AntonAndell  need to discuss APR estimation using additional tokens in rewards

hetfly:
@FezBox need to wait a bit before rewards can be estimated, then this can be tested

parrot9design:
For the APR tooltip:
Screenshot 2025-01-15 at 2 19 01 PM

> The rewards APR represents the USD value of rewards allocated to a pool, claimable from the Home page.
> 
> The fee APR represents swap fees earned by a pool over the past 30 days.

---

To avoid adding a third layer to the table when pools have multiple reward currencies, let's change the BALN/sICX APR to `Rewards`, and add a tooltip when multiple reward currencies are available:

Screenshot 2025-01-15 at 2 24 11 PM
Screenshot 2025-01-15 at 2 28 07 PM


> Paid in BALN (x – x%) and sICX (x%).
>
> Your BALN rate depends on the amount of bBALN you hold.

(If they have a position with staked LP tokens, show the specific BALN rate rather than the range.)

<br>

If BALN on ICON is the only reward currency, show this variation:

> Paid in BALN. Your rate depends on the amount of bBALN you hold.

---

Let's also change `Liquidity incentives` to `Liquidity rewards` in the Rewards panel on the Home page.
Screenshot 2025-01-15 at 2 04 56 PM

hetfly:
@parrot9design do you want to have the rewards tooltip next to every incentivised pair rewards? Seemed very messy when I tried it, can we show the tooltip only when the pool has both sICX and BALN rewards?

Also, we can't show a single number for BALN rewards even if user has a bBALN position, the boost is derived not only from bBALN amount but also from the amount of staked LP, which is shown in the `Your pools` section.

parrot9design:
Yeah, we can keep the range for the `Core pools` section. The mockup above was taken from the `Your pools` section which is why it's a single number – sorry for the confusion.

For those two tooltip variations, we can elaborate a bit more:

>Your BALN rate depends on your position size and the amount of bBALN you hold.

>Paid in BALN. Your rate depends on your position size and the amount of bBALN you hold.



Regarding your earlier point about too many tooltips, I have been torn about that, as well. It's necessary to explain somewhere why the range exists, but as pools are about to go cross-chain and bBALN is not, it didn't make sense to keep in the main APR tooltip. Open to feedback/alternatives for handling this.

hetfly:
@parrot9design the tooltip now shows for pairs with BALN rewards and BALN rewards with external rewards. Doesn't show for external rewards only. I think the tooltips are too much, though. We could use the the tooltip and `? Rewards` labeling in cases with multiple reward tokens only and explain the BALN range in the header tooltip. I wouldn't mind mentioning the bBALN is exclusive to ICON network, but I understand you want to omit `ICON` mentions anywhere possible. Still would lean towards explaining it in the header tooltip instead of having a tooltip next to every BALN reward range.

Screenshot 2025-01-16 at 13 03 09

<hr />

Note to myself, remove sICX dummy rewards

parrot9design:
Yeah, you're not wrong about `? Rewards` being too much. The idea was to keep the labelling consistent with `Rewards` instead of the specific token so that it's easier to talk about in the main tooltip (`The rewards APR represents...`), which meant it made less sense to explain the range there also. 

Let's:
- Use `? Rewards` when there are multiple currencies paid out
- Specify the currency when only one is available
- Update the main tooltip to say:

>Based on the USD value of liquidity rewards (claimable from the Home page) and fees earned by a pool over the past 30 days.
>
>BALN rewards depend on your position size and bBALN holdings.


I don't think we need to specify ICON until liquidity goes cross-chain. At that point, we can add `(boost available on ICON only)` to the end of the 2nd sentence.

hetfly:
There are currently no rewards for the wICX/sICX pool, but the PR is ready to merge whenever @parrot9design @FezBox

parrot9design:
- [ ] 4 decimal places might be a bit much for sICX?
Screenshot 2025-01-20 at 9 38 08 AM


- [ ] wICX appears in the liquidity rewards tooltip. Also unsure why it says I need bBALN when I can't boost my rewards?
Screenshot 2025-01-20 at 9 40 02 AM

hetfly:
>4 decimal places might be a bit much for sICX?

.0000 is hidden now, it's a side effect of the formatter we use, shows four decimals if the value/balance is below 0.01. 

<hr />

>wICX appears in the liquidity rewards tooltip. Also unsure why it says I need bBALN when I can't boost my rewards?

Good question. Votable pools with 0 BALN rewards are hidden from the tooltip now.

parrot9design:
It still says I need bBALN to maximise my return?
Screenshot 2025-01-21 at 10 32 22 AM
I think we can hide the liquidity rewards tooltip when there are no pools to display.


Aside from that, this looks ready to merge. Thanks @hetfly!

PR #1545, opened 2024-11-28
Feat/integrate balancedx

0xmilktea:


parrot9design:
Let's display the route for intent-based trades like this:
Screenshot 2024-11-27 at 8 23 56 AM

0xmilktea:
- Update fromChain type in the intents SDK.
- Remove ts-ignore

parrot9design:
Initial swap modal:


The `Swap` button should stay intact until you've authorised the transaction in your wallet.

Once you've authorised the transaction:



We should prevent the user from being able to click outside the modal until the order is complete or they cancel it.

Once complete, show this state for a few seconds before closing:


Or show this state if they cancel it:



Let me know if I've misunderstood anything about how this works. I've assumed that if the solver can't process the swap before a user gets impatient and closes out of the modal, it would be better to cancel the swap and start over than set aside space for it somewhere else in the UI. (We do want to revisit transaction handling in general, but that would take longer to implement than the time we have to get this release out the door.)

---

Let's add `Swap time` underneath the fee in the intent version of the dropdown:
Screenshot 2024-12-04 at 7 37 37 AM

As the route will be the from asset to the for asset, do we want to label the route for these ones `Balanced Intent`? That way it's clear that it's using a different option to trade altogether.

Curious what the price impact would be (if any) for intent trades?

parrot9design:
There's about a 10 second delay between confirming the transaction and the button changing to the `Swapping` state.
https://github.com/user-attachments/assets/4f4eb225-f3ba-4880-b38c-41ead9ae5e56


When a transaction shows that it's been completed, we should hide the `Cancel` & `Swap` buttons.
Screenshot 2024-12-06 at 8 25 34 AM

parrot9design:
- [ ] Let's remove the error message that appears if you cancel the transaction in your wallet. 
Screenshot 2024-12-06 at 4 05 49 PM

- [ ] Before we ship, can we update the default pair so it's not ETH > SUI? Better to keep it on a pair that supports a chart.

parrot9design:
Just to clarify, the lower limit is $100? I've found it doesn't switch over to the intent trades at a consistent value, and often switches over randomly, like it needs to wait a certain amount of time before it can process the next trade?

---

I got this error during my last couple of attempts:
Screenshot 2024-12-07 at 7 26 52 AM

Let's just say `Swap failed` and remove the message in red.

Clicking Swap in this modal triggers the same error, so I think we could just add a close button (rather than auto-hide the modal after 5 seconds) so people know the status of their swap.
Screenshot 2024-12-07 at 7 38 42 AM

---

After a successful transaction, I clicked the asset switcher to trade the inverse and an empty modal randomly appeared. This happened yesterday as well, but I wasn't able to capture it on video.

https://github.com/user-attachments/assets/fb4888af-5975-438f-9efe-acf084b633cb

---

I have not been able to test manually cancelling an order yet.

parrot9design:
For the error handling in the Swap panel:

- Remove the `Pending transactions` header (their presence should make that obvious)
- Bold the amounts in the transaction message to increase readability (e.g. **200 SUI** for **0.1 ETH**)

It should look as simple as this:
image

parrot9design:
After a successful swap, if I leave the UI open (e.g. while I look at a different app), an empty confirmation modal will always appear:
Screenshot 2024-12-09 at 7 16 06 PM

parrot9design:
Got these errors during my last round of testing:

Screenshot 2024-12-10 at 8 53 33 AMScreenshot 2024-12-10 at 9 14 13 AM


As the swaps appear under the Swap UI anyway, let's remove the errors from the modal so it looks like this (and then disappears after ~5 seconds like the Completed state):

slice1

---

When trying to cancel outstanding swaps, it wouldn't let me cancel the first one until I cancelled the second. 

https://github.com/user-attachments/assets/b92a15d1-ed91-4fb2-a954-ad315da3f68b

Is there anything we can do to prevent this looking so awkward?
Screenshot 2024-12-10 at 9 25 29 AM


---

I tried to reproduce the issue with the empty modal randomly appearing, but I've only managed to make it appear once, and not while recording. When it happened, I'd just finished an intent trade then swapped the pair around and clicked 100% for ETH.

parrot9design:
Managed to catch the random modal:

https://github.com/user-attachments/assets/ef64702e-f74d-46c0-a3e5-537f0f6103f8

But in this case, it was after a non-intent swap. 🤔

0xmilktea:
- add tx timer

parrot9design:
Let's change `Cancel` to `Close` here:
Screenshot 2024-12-11 at 8 42 17 AM

---

It took about 4 attempts before I was able to cancel a different swap. Everything was fine on the wallet side so I'm not sure what the issue was. I started the video after the second attempt failed.
https://github.com/user-attachments/assets/4c8842de-4c5a-4291-903c-d87f2d4db0e2

---

The empty modal appeared shortly after each of the above swaps failed. It also sometimes appears after a successful transaction, when the original swap modal closes without showing that it's been completed. 

---

We don't need to get this in for the initial launch, but can you share a list of the various error messages so I can rewrite them?

parrot9design:
The button to adjust your bBALN is missing:

Screenshot 2024-12-12 at 9 39 55 AM
Screenshot 2024-12-12 at 9 42 20 AM

Assume we just need to pull through the changes from yesterday's releases, but flagging here just in case.

PR #1600, opened 2025-01-21
Implement withdrawal limits notice

### LINKED ISSUE #1283: Add a message under the Swap & Transfer buttons when the withdrawal limit has been hit
Due to recent market activity, people have been more proactive on the exchange. This means they keep encountering the withdrawal limits and are confused when their transactions won't go through.

We should add a message to the Swap button when you enter an amount that can't be fulfilled at the current time.

- When the withdrawal limit has been hit, use `{ticker} unavailable for {amount}h {amount}m`:

Screenshot 2024-03-05 at 10 16 48 AM

<br>
<br>

- If the withdrawal limit hasn't been hit, but you enter an amount that will take you over it:

>More {ticker} available in {amount}h {amount}m

If the logic is a hassle, we can skip this variation, I just wanted to make the differentiation that _some_ of the asset is still available.

Also, if we're unable to fetch the time remaining before the limits reset, we can use `{ticker} is temporarily unavailable` as an alternative (although it's far less useful).

---

While we're focusing on this piece, there's one other issue I'd like to address if possible. When you enter any number into a swap field, the button label initially changes to `Insufficient liquidity` while it calculates the trade. Is it possible to prevent it from appearing here?

hetfly:
close #1283

### DIFF
--- a/apps/stats/src/sections/TokenSection.tsx
+++ b/apps/stats/src/sections/TokenSection.tsx
@@ -370,18 +370,7 @@ export default React.memo(function TokenSection() {
               >
                 Available
               </HeaderText>
-              <HeaderText
-                style={{ cursor: 'default' }}
-                // role="button"
-                // className={sortBy.key === 'holders' ? sortBy.order : ''}
-                // onClick={() =>
-                //   handleSortSelect({
-                //     key: 'holders',
-                //   })
-                // }
-              >
-                7d trend
-              </HeaderText>
+              <HeaderText style={{ cursor: 'default' }}>7d trend</HeaderText>
             </DashGrid>
           )}
 
--- a/apps/web/src/app/components/AssetManagerTokenBreakdown/index.tsx
+++ b/apps/web/src/app/components/AssetManagerTokenBreakdown/index.tsx
@@ -2,7 +2,7 @@ import { Typography } from '@/app/theme';
 import { useRatesWithOracle } from '@/queries/reward';
 import { formatBalance } from '@/utils/formatter';
 import { CurrencyAmount } from '@balancednetwork/sdk-core';
-import { xChainMap } from '@balancednetwork/xwagmi';
+import { xChainMap, xTokenMap } from '@balancednetwork/xwagmi';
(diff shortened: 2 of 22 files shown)


hetfly:
@parrot9design We can't show the dynamic amount of the time remaining for the reset. The reset is always linear over the limit period and is reset with every withdrawal. But the user can at least see the amount in `Only {amount} available` change every couple of seconds, if it's being topped up.

<hr />

The warning for lower limit in the token section is setup to trigger when half of it is depleted, same as it's on the stats page to keep it consistent, but we can update it on both places, if half seem too high.

image

parrot9design:
- [ ] Remove the tooltip next to `liquidity`
Screenshot 2025-01-30 at 4 32 58 PM

- [ ] Order the blockchains from largest to smallest amount
Screenshot 2025-01-30 at 4 34 16 PM

- [ ] Consider using the specific token names for the BTC variants (but formatted better than I have here):
Screenshot 2025-01-30 at 4 40 02 PM

- [ ] Adjust the Available tooltip to avoid hardcoding the withdrawal amount:

>To protect liquidity, many assets include a withdrawal limit that resets over 24 hours.

- [ ] Consider adding the withdrawal % next to the asset's available value, just like we do on the Stats page.
Screenshot 2025-01-30 at 5 00 28 PM

parrot9design:
This looks good to merge, thanks @hetfly! 


A couple things I discovered that we can (and probably should) address in a separate PR:

- Should we remove Archway from the USDC entry throughout the Trade page? The Stability Fund doesn't accept Archway USDC so you can't trade into it, and I assume if you try to swap USDC from Archway, the transaction will fail (though I don't have any to test).
Screenshot 2025-01-31 at 11 45 41 AM
Screenshot 2025-01-31 at 11 45 16 AM

- If the slippage is too high, the button label used to say `Reduce price impact`. This no longer appears – assuming it was lost during the changes for intent-based trades. Are we able to fix that so there's more context as to _why_ you can't proceed with the swap?
Screenshot 2025-01-31 at 11 50 25 AM

parrot9design (approved):

PR #1603, opened 2025-01-24
Check Solana account existence before bridging; require >= 0.002 SOL for new accounts

swiftcc624:


parrot9design:
- [ ] Let's revert to `Swap`/`Transfer`for the button label, unless another message e.g. `Insufficient liquidity` applies.

- [ ] For the SOL message, let's use `0.002 SOL required for new Solana wallets.` If you can send other assets to the wallet without any issue, it's not really an activation problem.
Screenshot 2025-01-28 at 12 55 00 PM

- [ ] As it only affects SOL, let's also make the value clickable on the Bridge tab (previously assumed you'd need to go to the Swap tab to resolve it).

swiftcc624:
>> As it only affects SOL, let's also make the value clickable on the Bridge tab (previously assumed you'd need to go to the Swap tab to resolve it).

- why does it goto swap tab from bridge tab?
- if go to swap tab, what is the input token of the swap pair? we know output token is SOL

parrot9design:
Ignore me, just reflecting on why I didn't pursue that direction previously.

We just need to handle it the same way as the Swap tab: make `0.002 SOL` a link which fills that amount for SOL on the Bridge tab, and make the `Transfer` button inactive until the amount is 0.002 SOL or higher.
Screenshot 2025-01-28 at 7 55 15 PM

swiftcc624:
@parrot9design I've fixed it. can you please check?

parrot9design:
Let's make the `Transfer` button inactive when the message appears in the signed-out state: 
Screenshot 2025-01-28 at 8 21 54 PM

Otherwise, it's good to ship. 👍

swiftcc624:
@parrot9design I fixed it, can I merge it after deployed?

parrot9design:
Looks good to me, feel free to merge.

PR #1627, opened 2025-02-06
fix: regressions on production

0xmilktea:


parrot9design:
Looks good, thank you. 

Just need to update the button in the Bridge panel from `Enter amount` to `Transfer` (active) in the signed out state:

Screenshot 2025-02-12 at 10 57 49 AM

Then it's ready to merge.

parrot9design:
@0xmilktea one more thing we could tag onto this PR:

image

PR #1638, opened 2025-02-13
Support currency filter by chain

hetfly:


hetfly:
Note to self, remove intent staging url

parrot9design:
Looks good, thanks @hetfly! Let's reduce the font size in the button to 14px, then this is good to merge. 💪

PR #1665, opened 2025-03-21
Add Stellar trustline creation

### LINKED ISSUE #1661: Adding Trustline in Stellar Account for holding Assets
Similar to creating an account to hold XLM, adding a trustline for each asset is required to hold other assets. This can be done in two ways: either through sponsorship or by the user creating it themselves. Since users are likely to already have XLM before acquiring USDC, it would be more practical for them to create the trustline on their own than sponsorship. 

When transferring cross-chain assets (USDC for now) to Stellar, we can check whether the trustline for the destination account exists. If not, the frontend can prompt the user to create one and facilitate the process.

This is sample reference code to check if trustline exists and create trustline for an account for asset.

```javascript

const checkIfTrustlineExists = (balances, asset_code, asset_issuer) => {
  return balances.some(
    ({ asset_code: code, asset_issuer: issuer }) => code === asset_code && issuer === asset_issuer
  );
};

async function createTrustlineIfNecessary(accountId, asset_code, asset_issuer, secretKey) {
  try {
    const server = new SorobanRpc.Server("https://soroban-testnet.stellar.org");
    const horizonServer = new Horizon.Server("https://horizon-testnet.stellar.org");

    const sourceAccount = await server.getAccount(accountId);
    const { balances } = await horizonServer.accounts().accountId(accountId).call();

    if (checkIfTrustlineExists(balances, asset_code, asset_issuer)) {
      console.log("Trustline already exists");
      return;
    }

    const transaction = new TransactionBuilder(sourceAccount, {
      fee: BASE_FEE,
      networkPassphrase: Networks.TESTNET,
    })
      .addOperation(Operation.changeTrust({ asset: new Asset(asset_code, asset_issuer) }))
      .setTimeout(180)
      .build();

    transaction.sign(Keypair.fromSecret(secretKey));

    const result = await server.sendTransaction(TransactionBuilder.fromXDR(transaction.toXDR(), Networks.TESTNET));
    await wait(result);
  } catch (error) {
    console.error("Error in creating trustline:", error.message);
  }
}

```

Mainnet USDC
asset_code: USDC
asset_issuer: GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN
contract_id: CCW67TSZV3SSS2HXMBQ5JFGCKJNXKZM7UQUWUZPUTHXSTZLEO7SJMI75

Testnet USDC
asset_code: USDC
asset_issuer: GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5
contract_id: CBIELTK6YBZJU5UP2WWQEUCYKLPU6AUNZ2BQ4WWFEIE3USCIHMXQDAMA

hetfly:
close #1661


parrot9design:
- [ ] Let's change this button label to `Validating Stellar wallet`: 
Screenshot 2025-03-28 at 11 26 56 AM


- [ ] Let's make sure the `Swap` button remains inactive until USDC has been activated:
Screenshot 2025-03-28 at 11 36 15 AM

- [ ] Change the error message to `You need at least 1.5 XLM in your Stellar wallet.`. 
Screenshot 2025-03-28 at 11 37 00 AM

Also, why do they need so much XLM in the wallet when a swap from Stellar only costs 0.07 XLM in gas?


- [ ] I tried to test the full flow with a new Stellar wallet (no XLM balance), but I can't proceed beyond the initial activation transaction, which remains in the pending state with the button still active.
Screenshot 2025-03-28 at 11 44 00 AM

hetfly:
>Also, why do they need so much XLM in the wallet when a swap from Stellar only costs 0.07 XLM in gas?

That's a Stellar thing. An account needs > 1 XLM to be considered active and it also needs 0.5 XLM for a trustline, which is basically allowing the asset in your wallet. So nothing to do with the fees.

hetfly:
>I tried to test the full flow with a new Stellar wallet (no XLM balance), but I can't proceed beyond the initial activation transaction, which remains in the pending state with the button still active.

That's a CORS error on the deployment URL. You can activate the wallet in the live app and then go back to this PR.

parrot9design (approved):

PR #1694, opened 2025-05-01
Add stellar intents

hetfly:


hetfly:
Note to self, remove `isMMBetter=true`

parrot9design:
I'm getting this error every time I try an intent trade with Stellar:
Screenshot 2025-05-07 at 2 28 56 PM

hetfly:
@parrot9design I can't replicate the issue. Is it still the case for you?

Note that intent trades are forced now for testing purposes, other swaps won't work, but you can test the stellar ones with almost any amount, lower limit is around $0.1

Also note that if the swap tx will go through, sometimes they get timed out. It's already reported

parrot9design:
>I can't replicate the issue. Is it still the case for you?

No, I didn't encounter this during testing today. Must've been one of the issues Alex fixed over the weekend.

Just to verify, this is the error message that appears when the tx times out? Or am I hitting a different issue?
Screenshot 2025-05-13 at 4 23 45 PM

And trades not supported via intents are appearing like this, rather than the button showing `Swap not supported` as it would normally. Assume this is expected due to the testing adjustments?
Screenshot 2025-05-13 at 4 22 56 PM

Many transactions took 20-45 seconds to complete, but I assume this will be addressed with the new solver. Otherwise, all looks good.

hetfly:
@parrot9design timeout limit was increased, solver bug fixed, Alex is fine with releasing.

>Assume this is expected due to the testing adjustments?

Yes and it's removed now.

Let me know if we're ok to merge 🙏

parrot9design:
I got this error a few times after a swap had been pending for 40-45 seconds:
Screenshot 2025-05-15 at 9 03 52 AM Screenshot 2025-05-15 at 9 09 33 AM
When I tried the exact same trades again afterwards, they went through without any issues. Though the average trade time was 20-45 seconds.

A tBTC > Stellar USDC intent took more than 1 minute, so I tried to cancel it. This appeared to work except that the tBTC never returned to my wallet. I then went to https://app.balanced.network/intent-check and successfully cancelled the order there, but still nothing. If I recheck the tx hash (`0x3bbb1164a94d2cee965356ece2f1791c9856795d71b401aaa3162743fe566532`), it gives me the option to cancel it again.

hetfly:
Thank you @parrot9design 🙏  

The timeout issue is on the SDK side and is now being worked on.

hetfly:
@parrot9design after updating the sdk with priority fee I have not encountered any swap timeouts. However it does increase the cost quite a bit so we had to increase the gas threshold, as well. It's 3 XLM now.

parrot9design:
Thanks! I also didn't encounter the timeout issue, but I could only test swaps between Stellar & Polygon because most of the intent-only tokens are missing. None of the SUI LST tokens, tBTC, wstETH, or weETH appear in the `All assets` list. 
Screenshot 2025-05-17 at 12 57 53 PMScreenshot 2025-05-17 at 1 13 57 PMScreenshot 2025-05-17 at 12 58 40 PM

hetfly:
@parrot9design that's weird. I can see them all, maybe a temporary fluke, can you check again please?

parrot9design:
I can see all the tokens again now, too. Must've been related to the issues with the pyth oracle over the weekend.

Intents seem to be working well now, so I think we can ship this whenever you're ready.

---

Separately, I did stumble across an issue with Sui which I'm also seeing in the live app. The balance of Sui assets keeps changing, as you can see in this video:

https://github.com/user-attachments/assets/5b4682ce-70e4-4bd4-8801-ec26796aaaf8

Because of it, I'm unable to remove my SUI collateral, nor can I trade any assets from Sui (hangs on the `Signing` message in Hana), even when it displays the correct balance.
Screenshot 2025-05-19 at 7 53 19 PM

PR #1708, opened 2025-05-30
Improve stellar transaction UX

hetfly:


hetfly:
@parrot9design as per MJ reporting stellar transaction failing and not closing the confirmation modal afterwards, I've made a change in handling stellar transactions that lead to higher delays before showing `Swapping` in the button and a `Confirming on...` message. To handle for this change the swap, bridge and loan buttons are disabled (greyed out) after clicking them once and prompting the wallet. 

Have a look and let me know if you're fine with it please

parrot9design:
Hmm, the only problem with this approach is that if you close the wallet without signing the transaction (or the wallet doesn't appear at all), the button doesn't become active again. Even after closing the modal, adjusting e.g. the amount of collateral to withdraw and then trying again, the modal button remains inactive. I have to refresh the page before I can take action.

https://github.com/user-attachments/assets/6ea624d5-6f85-4c80-91ae-685094ef1395

Because of the delay before the modal changes, I think we need to do something else with that button, like add the spinning logo after the label or update the text, so people don't think that it's stuck or something. I'm not coming up with any good solutions right now, but I'll continue to think that over. 

Are you able to pull in the updates from master so I can test it alongside the improvements from #1709?

hetfly:
@parrot9design master merged, will tackle the rest later

hetfly:
@parrot9design I've also merged https://github.com/balancednetwork/balanced-network-interface/pull/1713 in here, that should take care of the ICON intent bug, finally. It changed the execution flow a little bit, but from my testing all looks good.

PR #1718, opened 2025-07-10
Feat/add sodax solver

hetfly:


parrot9design:
## Recent activity

- [ ] `Recent Activity ` > `Recent activity`
- [ ] Replace empty state text with `No recent trades to display.` since it doesn't cover all transaction types for now.

Screenshot 2025-08-27 at 11 31 51 AMScreenshot 2025-08-27 at 11 32 46 AM

- [ ] Let's also replace the icon with this one: 
anti-clock
[anti-clock.svg.zip](https://github.com/user-attachments/files/21998273/anti-clock.svg.zip)


---

## Swaps

So, Injective won't be included in the release, right? Still able to set up swaps from Injective, but nothing happens when I click Swap in the confirmation modal.
Screenshot 2025-08-27 at 11 43 56 AM

I can receive funds on Injective, though a swap from Sui USDC > INJ took 1:10 after signing the tx.

--

- [ ] We're not keeping this section, right? 😅

Screenshot 2025-08-27 at 11 46 03 AM

--

- [ ] If you click on the wallet or Recent Activity icon to close the dropdown, the colour doesn’t revert back to its unclicked state:

Screenshot 2025-08-27 at 11 50 49 AM

--

- [ ] There's no reason to keep the slippage tolerance section now, right?

Screenshot 2025-08-27 at 11 56 02 AM

--

- [ ] Update the swap & cancel transaction flows based on the designs I shared separately.

--

- [ ] When you enter an amount for a swap, the button flashes to `Swap not supported` before the solver returns a quote. Technically all swaps will be supported now, right? Can we change it to a loading animation or something instead?

- [ ] When the swap legitimately isn't supported because the amount exceeds the maximum, can we update the label to use one of these variations instead? Either `Swap ${amount} or less` if we know the maximum swap size, or `Reduce swap amount` if we don't. 

--

- [ ] After you approve an asset, it can take a while for the button to change to Swap, so I've signed multiple approve tx for the same swap. I waited 1:11 for USDC on Avalanche, but it only changed to Swap after I switched to a different app window and came back again.
Screenshot 2025-08-27 at 1 20 50 PM

--

- [ ] I didn't realise this swap had failed until I opened the Recent Activity dropdown:
Screenshot 2025-08-27 at 1 29 21 PM

Can we draw attention to it while it's still within the 5 min deadline? Whether we make the RA circle red as I included in the pending tx mockups, show a `x` icon instead of the tick, or do something else like a red notification badge. (A combination of the 2nd and 3rd options might make sense given that people could have more than 1 trade in progress at a time.)

Also, it didn't seem to cancel on its own after 5 minutes as I expected.
Screenshot 2025-08-27 at 1 37 14 PM

I assume that's on the SODAX side? If not, we'll need a better way to handle cancellations for people who start a new session before trying to cancel manually, as all transactions are wiped when you close the browser.

--

- [ ] Unable to make any swaps from Stellar atm. Not sure if that's expected?
Screenshot 2025-08-27 at 1 00 54 PM Screenshot 2025-08-27 at 12 44 49 PM

Also, are we able to adjust what the error messages say? Or is that out of our hands? The one that appears if you cancel in the wallet would be nice to remove entirely (because the swap didn't fail, you just decided to go back).

--

- [ ] BALN should have the ICON logo attached since it's only available to trade on one chain.
Screenshot 2025-08-27 at 11 37 54 AM

- [ ] BALN also doesn't appear in the asset list if you have the ICON chain selected:
Screenshot 2025-08-27 at 1 54 50 PM

--

- [ ] We can remove Archway & Havah from the sign-in modal now, right? Or does that have to stay in place because of the `/trade-legacy` page?

--

- [ ] Replace the community token list with a redirect to `/trade-legacy`:

Screenshot 2025-08-27 at 2 04 54 PM

We may also want to keep it visible at the bottom of the list so you don't have to type to see that it's an option.

Not sure if there's anything else we can/should do to bring people's attention to the assets they hold that are no longer supported on the main Trade page.  🤔 Will continue to mull it over, but let me know if you have any ideas.

---

## Legacy trade page

- [ ] Add `Legacy` to the Trade page title to differentiate between the two versions:

Screenshot 2025-08-27 at 1 57 17 PM

- [ ] If you click the Trade navigation item once on the legacy page, you remain there. It should probably take you to the proper Trade page?

hetfly:
Just an update, still work in progress

<hr />


>So, Injective won't be included in the release, right? Still able to set up swaps from Injective, but nothing happens when I click Swap in the confirmation modal.


I was wondering if you'd managed to do another swaps as you mentioned on the call 😅 
Injective is now disabled

<hr />

>We're not keeping this section, right? 😅

Just for an easy access when debugging and sharing info with Alex, will remove it once we're done testing

<hr />

>There's no reason to keep the slippage tolerance section now, right?

There is actually. The slippage allows for faster intent execution, it sets the minimum user is willing to receive, the order then gets executed at best possible price, but not below the minimum stated, as fast as possible.

<hr />

>When the swap legitimately isn't supported because the amount exceeds the maximum, can we update the label to use one of these variations instead? Either Swap ${amount} or less if we know the maximum swap size, or Reduce swap amount if we don't.

Currently not possible because of the error solver returns. It will be improved in the future. 

<hr />

>Also, it didn't seem to cancel on its own after 5 minutes as I expected.

This is an unpleasant issue with the solver now. We cannot derive cancel status from it as of now. (they are working on improving status management) So the only way we could reliably show cancel status is when user cancels manually (and we check the actual response from the sdk function, not the order status from the solver). 

We could also show cancelled automatically if the order is older then the deadline, but that is not very reliable. 

How it functions now is you can try to cancel the order even after the deadline and if it's cancelled, it updates the state in the UI (based on the error it receives from the sdk when you try to cancel non existing order) or it let's you cancel the order if it's still not cancelled.

>we'll need a better way to handle cancellations for people who start a new session before trying to cancel manually, as all transactions are wiped when you close the browser.

they are not wiped, they persist through the sessions, however you tested it on two different URLs, each deployment has a specific URL and thus it's own local storage key.

<hr />

>We can remove Archway & Havah from the sign-in modal now, right? Or does that have to stay in place because of the /trade-legacy page?

I think we should leave it in place in this version, there might be some people that still use Balanced from these chains, at least Havah I think will have some activity

parrot9design:
Thanks for the update. 🙏

- [ ] Instead of `Loading...` for the button label, let's try `Finding best price...`

- [ ] For the Recent Activity animation, maybe just spin the outer circle and keep the clock hands static? We can provide a better icon if this one doesn't allow that much flexibility.

- [ ] Let's update the Recent Activity transaction design a little. I had a couple failed transactions, but I wouldn't know it when clicking on one because it only links to the first part of the transaction, which succeeded. This variation allows us to link to the tracker for both the asset sent and received, and gives the icons a little more space to breathe.
recent-activity-mockup

Here's the arrow: [swap-arrow.svg.zip](https://github.com/user-attachments/files/22017317/swap-arrow.svg.zip). Though if you want to hold off on this, Pete will add some code to the demo tomorrow to make the styles a bit easier for you.

Also, if the time is 1 minute ago, can we remove the `s` so it's `1 min ago` instead of `1 mins ago`?

hetfly:
> Update the swap & cancel transaction flows based on the designs I shared separately.

Did what I could but `Waiting for wallet...` is outside our control now. With these sdks the first response the app gets is order data. So we would skip the `Creating order...` step and go from `Waiting for wallet...` to `Order created` or a failure notice.

Also I think it's better to close the modal when the order is created and keep the context within the activity section and its transition states. Swaps take around 30secs which is too long for me to stare at the modal 😅 

<hr />

>I didn't realise this swap had failed until I opened the Recent Activity dropdown

Added alert background to the recent activity icon when there are failed transactions (meaning user can cancel from the activity section dropdown), and added popup notifications - those will probably need a content update 😅 

<hr />

>Unable to make any swaps from Stellar atm. Not sure if that's expected?

Is it still an issue for you? Worked fine on my end. There were some solver upgrades recently, you might have caught it during a downtime.

image

<hr />

>Also, are we able to adjust what the error messages say?

Unfortunately not

hetfly:
>Update BALN in the asset selector to use the ICON logo & appear under the ICON filter

BALN is currently not supported by the solver and is now taken out from currency dropdown.

hetfly:
> Bridged USDC to Solana, but it hasn't updated the wallet balance. Also, if I rotate the swap, the wallet balance disappears (we've had this issue in the past when the amount is 0).


> Tried to withdraw liquidity on Solana, but it won't trigger the wallet. I assumed this was related to the recent hack, but I have no issue when I attempt the same action in the live app.


I've changed the RPC, hope this solves the issue. Worked well for me on the latest deployment.

<hr />

> The Approve lag is a constant for every trade that requires it. Can we do anything about this (or ping the SODAX devs to address it in the SDK)? I can't get it to change to Swap unless I swap between spaces multiple times on my Mac. Can reproduce with Hana, Rabby, and Metamask.

I have no issues with approve, but the app is now refetching the allowance data every second which should take care of your problem.

hetfly:
Went through the other comments and these are the outstanding tasks

- [ ] For the Recent Activity animation, maybe just spin the outer circle and keep the clock hands static? We can provide a better icon if this one doesn't allow that much flexibility.
- [ ]  For the asset received, is it possible to link to the transaction on the destination chain instead? We'll have to wait until that transaction is available before we can embed it, but I'm okay with that because the Sonic explorer is so hard to decipher.
- [ ] After signing a cancel order transaction from Stellar, the modal kept reverting back to the Cancel state (it still showed as pending in the Recent Activity dropdown, in case that helps)

<hr >

Additionally 

- [ ] Migrate panel for bnUSD
- [ ] Update content on the homepage to inform user about old/new bnUSD situation

also, do we use the `Migrate` panel for ICX and BALN migrations, as well ?

parrot9design:
>also, do we use the Migrate panel for ICX and BALN migrations, as well ?

Discussed this yesterday but just to address it here: we'll only support the bnUSD and ICX migrations to start with. BALN will come later, after the devs finish the SODA staking work and we have a plan in place to halt governance & BALN inflation and allow people to unlock BALN without a penalty fee.

---

- [ ] When I try to change the active blockchain via the blockchain dropdown, I can't choose Optimism for USDT or BNB Chain for USDC. However, I can choose them if I click the chain from the asset dropdown.

parrot9design:
I know you're still working on this, just dropping in some feedback.

- [ ] Make `bnUSD (old <> new)` the default option on the Migrate tab
- [ ] Add a chevron after the active migrate option to make it more apparent that there are more options to choose from
- [ ] When used as a ticker, let's remove the space from bnUSD (old) so it's `bnUSD(old)` (we can always reduce the size/opacity of the `(old)` piece if needed
- [ ] Update the description for the bnUSD migration:
Screenshot 2025-09-06 at 12 07 31 PM

> Balanced is transitioning to a new technology stack, powered by [SODAX](https://www.sodax.com/).
> 
> - Use bnUSD to trade on every chain except ICON.
> - Use bnUSD(old) to repay loans and earn through the Savings Rate.

- [ ] Update the description for ICX <> SODA:
>The ICON blockchain is now [SODAX](https://news.sodax.com/posts/introducing-sodax), a unified liquidity layer and DeFi platform on Sonic. Swap your ICX 1:1 for SODA on the Sonic blockchain.
- [ ] Can we make it possible to open the legacy exchange in a new tab (not by default, but if a user wants to)? I was sad to see my usual command+click didn't work for this.
- [ ] For the swap route, let's use `SODAX Intent`:
Screenshot 2025-09-06 at 12 09 42 PM

- [ ] The wallet balances here should be `bnUSD(old)` for full clarity:
Screenshot 2025-09-06 at 12 13 38 PM

- [ ] My balances for bnUSD(old) aren't being detected in the Savings Rate section:
Screenshot 2025-09-06 at 12 13 00 PM
^That text string (Buy or borrow bnUSD...) should also say `bnUSD(old)`.

hetfly:
>What if you have bnUSD(old) on chains other than ICON, Sui, or Stellar? Can you just not do anything with it now? I thought we'd still allow people to migrate from the old version on any chain to the new version, but only allow the inverse on the 3 chains mentioned? Or is the expectation that they use to the legacy exchange to dispose of it on other chains?

Yeah, you need to legacy trade or bridge them.

<hr />

>On the legacy exchange, I received bnUSD(old) but the tx remained pending in the UI:

Unfortunately this is an xCallScanner issue and I doubt anyone will spend time on this. It's been known for a long time and priorities are elsewhere now.

parrot9design:
There have been some issues with the solver today so I haven't been able to test any swaps. Hopefully the issues below capture everything that's left to take care of. At this stage we're expecting to launch Tuesday morning your time.

---

- [ ] Let's add a bit of extra guidance to the Migrate tab:

>If you hold bnUSD(old) on a chain other than ICON, Stellar, or Sui, you'll need to transfer it via the [legacy exchange](https://app.balanced.network/trade-legacy/bridge) before you can migrate.
Screenshot 2025-09-08 at 10 51 52 AM

---

- [ ] If you click the Migrate button when signed out, surface the prompt to sign in instead of the confirmation modal.
image

---

- [ ] Position the arrow in this link right after `exchange` so it doesn't appear all the way to the right if it breaks across two lines (which it seems to do on my iPad but not my iPhone, for some reason).


---

- [ ] ICX > bnUSD(old) still supports a chart for some reason. Until you enter a value, the last price is just a string of zeroes. As far as I can tell, no other pair still supports a chart, so we can probably remove it until we add the new charts.
Screenshot 2025-09-08 at 11 38 27 AM

---

## Outstanding issues (repeated here for easier viewing)

- [ ] When I try to change the active blockchain via the blockchain dropdown, I can't choose Optimism for USDT or BNB Chain for USDC. However, I can choose them if I click the chain from the asset dropdown.
- [ ] After signing a cancel order transaction from Stellar, the modal kept reverting back to the Cancel state (it still showed as pending in the Recent Activity dropdown, in case that helps)  – note that I haven't been able to test this since I first raised it
- [ ] Add a chevron after the active migrate option to make it more apparent that there are more options to choose from (like the collateral selector).

- [ ] For the Recent Activity animation, maybe just spin the outer circle and keep the clock hands static? We can provide a better icon if this one doesn't allow that much flexibility.
- [ ] For the asset received, is it possible to link to the transaction on the destination chain instead?

parrot9design:
>>Do we want to hide the charts on the legacy exchange, or just the new one? They're probably still somewhat relevant for the legacy exchange, right? If not, we should have the "no price chart supported" message to prevent it from looking so empty.
>
>I'm for hiding them completely until resolved properly.

All good. Let's just add `No price chart available for this pair.` to the panel on the legacy exchange so it doesn't look as empty. 

---

- [ ] There's a problem using bnUSD(old) on Stellar (both with the migration tool and swaps/transfers on the legacy exchange). Raised in Discord but it shouldn't be related to the trustlines given that I already had bnUSD(old) in my Stellar wallet, so not sure what we can do on our side.

---

- [ ] Is there any point including the new bnUSD on the legacy exchange?
Screenshot 2025-09-10 at 10 05 57 AM

parrot9design:
We can add this post-launch if easier (not sure if/when the SDK will make the logic/detection easier), but we should handle for Solana's token creation fee so people aren't surprised to receive less than quoted when they receive a Solana token for the first time (like we were when we tested bnUSD a few days ago).

If we can factor the fee into the total we display, it can be just another fee presented in the confirmation modal:

>Solana token fee (included): {amount} {token}
Screenshot 2025-09-13 at 6 45 18 PM

If we can't factor the fee into the total, let's go with this option instead:

>Solana will deduct a "new token" fee of {amount} {token}.
Screenshot 2025-09-13 at 6 43 01 PM

hetfly:
>If we can't factor the fee into the total, let's go with this option instead:

@parrot9design I think I can only check the current balance of the token on Solana, so the warning is displayed if there is 0 balance of that token. Also updated the text to `can`, let me know if that's fine by you. I think it's better not to subtract it from the swap output in this case since we don't know if that fee will actually be taken.

PR #1724, opened 2025-09-15
feat: add icx soda migration

hetfly:


parrot9design:
Leaving all issues related to this + the main intents release here, so we can test and release everything once it's stable.

Below are all the ones I've discovered during my testing session this afternoon. Hopefully that's the last of them. 🤞 

## Breaking issues

- [ ] Investigate the white-screen issue for active loans. Not sure if this was due to the intents release or just unfortunate timing, but I haven’t encountered it in this PR (though I only started testing after the live app randomly started working again this afternoon)
    - If related to the release, check the bnUSD(old) price data, which was unavailable throughout the app until it fixed itself.
    
- [ ] Couldn’t stake LP tokens for ICX/sICX (didn’t receive the prompt after supplying, and while I could adjust the stake through the expanded view, the confirmation modal wouldn’t trigger the wallet).

- [ ] Could stake LP tokens for other pools when prompted (tried with ETH/bnUSD), but the staking slider didn’t appear in the expanded view afterwards so I couldn’t unstake or withdraw liquidity (shows the message about staking via a third-party app instead).
    - I tried this in the live app before we reverted the release. When I checked the live app this afternoon, it acted like I hadn’t staked any LP tokens. So I staked there, then referenced this build, and again, saw the third-party message.

- [ ] Found another white-screen issue when testing this PR that can appear if you use the Swap tab: if you enter a value of `1000` or above into the second swap field for ICX, POL, S, bnUSD, or bnUSD(old), it trips out.

---

## Non-breaking issues
Listed in rough order of importance. Would be nice to tick off the first 2 before we launch.

- [ ] Update the price data for the S token, as it's quite far off. I thought the solver just couldn’t source a good rate, but currently, 1 S = $0.21 in the UI while it’s valued at $0.299 on TradingView


- [ ] Add SODA to the wallet dropdown.

- [ ] POL on the legacy exchange should still support tokens held on ICON, right? It also doesn't seem to have a price in the input field or the asset selector.
Screenshot 2025-09-16 at 4 32 04 PM

- [ ] I have more bnUSD(old) than any other asset in my wallet, yet the legacy exchange lists it last in the asset selector when sorted by wallet value:
Screenshot 2025-09-16 at 4 39 43 PMScreenshot 2025-09-16 at 4 39 49 PM


Update some visual bugs with bnUSD(old) on the legacy exchange:

- [ ] Liquidity positions wrap to two lines but break awkwardly to do so (maybe just add a space on either side of the `/`?)
Screenshot 2025-09-16 at 8 55 14 AM

- [ ] bnUSD(old) ticker in the Tokens list is a bit awkward (maybe change the name to `Legacy bnUSD`? I’ve verified that it won’t be funky at any breakpoint)
Screenshot 2025-09-16 at 8 40 52 AM
Screenshot 2025-09-16 at 2 50 29 PM

- [ ] For some reason, the asset order for LPs you’ve supplied to keeps flipping for anything that uses bnUSD(old)
Screenshot 2025-09-16 at 8 55 35 AM

(left this one till last as I know it'll be a pain to fix, up to you if you think it's worth tackling now, later, or at all)

PR #1732, opened 2025-09-26
Feat/add new price charts

hetfly:


parrot9design:
- [ ] The line chart doesn't take up the full width for the short time frames, anything we can do to fix that?
Screenshot 2025-09-30 at 9 58 55 AM Screenshot 2025-09-30 at 9 59 04 AM

---

- [ ] Let's switch the order so `Candles` comes before `Line`, since it's the default and also first alphabetically.

---
- [ ] When the same token is used in both input fields, we should just show the one option on the chart:
Screenshot 2025-09-30 at 10 17 31 AM
Screenshot 2025-09-30 at 10 17 46 AM

---

- [ ] Shouldn't bnUSD(old) have a price and chart? As the data is (currently) the same for both versions, we could also just show `bnUSD` when bnUSD(old) has been selected as one of the input tokens.
Screenshot 2025-09-30 at 10 04 15 AM

- [ ] afSUI, haSUI, vSUI, S, weETH, and wstETH have price data on Coingecko, so we should show the current price and chart for them also.
- [ ] mSUI doesn't have price data on Coingecko, but we should still show the current price since we use it elsewhere, and update the `Chart not available for {ticker}` message to end with a `.`.
Screenshot 2025-09-30 at 10 09 06 AM

---

- [ ] Some of the values on the y axis aren't being displayed properly. The line chart often shows 0-1 value along the side (particularly for stablecoins), and the price shown on hover often doesn't provide enough precision – based on the padding, I think it's being cut off.
Screenshot 2025-09-30 at 10 21 50 AM
Screenshot 2025-09-30 at 10 22 01 AM
Screenshot 2025-09-30 at 10 21 43 AM

---

- [ ] Trying to reveal the price at a specific point on the chart is quite difficult to do on a touch screen. You have to tap very intentionally, and you can't move along the line/candles like you can with a mouse hover on desktop. Not sure how much flexibility you have with the chart infrastructure, but is it possible to make it more seamless to reveal the data like the charts on the Stats page?

---

- [ ] The `Month` candle chart is quite dense compared to the others. Let's adjust the candlestick periods to make them all look relatively similar.

We can either replicate the density of the yearly view (visually appealing, but uses non-standard intervals):
- Week: 2 hours
- Month: 8 hours
- 6 months: 2 days
- Year: 4 days (no change)

Or replicate the density of the monthly chart (uses standard intervals, but not visually appealing):
- Week: 1 hour
- Month: 4 hours (no change)
- 6 months: 1 day
- Year: 2 days

Of the two, I naturally prefer the former, but I assume the non-standard intervals will require more aggregation work so I'll let you make the final call.

hetfly:
Chart for sonic is bugged for some weird reason, it doesn't show correct data, even when I try the name before rebrand.

hetfly:
>Some of the values on the y axis aren't being displayed properly. The line chart often shows 0-1 value along the side (particularly for stablecoins), and the price shown on hover often doesn't provide enough precision – based on the padding, I think it's being cut off.


This was so fkin painful 😅  
Came up with showing the highlighted current price which helps with spacing out the Y axis ticks but messes up the transition animations of the charts sometime. Let me know what feels better to you, can be compared with this deplpoyment:
https://balanced-network-interface-dui76wtz5-balanced-defi.vercel.app/

<hr />

>The Month candle chart is quite dense compared to the others. Let's adjust the candlestick periods to make them all look relatively similar.


Yearly queries were often failing. It's now replaced with the `Max` timeframe which works quite good, and uses monthly candles instead.

hetfly:
>Trying to reveal the price at a specific point on the chart is quite difficult to do on a touch screen. You have to tap very intentionally, and you can't move along the line/candles like you can with a mouse hover on desktop. Not sure how much flexibility you have with the chart infrastructure, but is it possible to make it more seamless to reveal the data like the charts on the Stats page?


Still having issues with this. The charts you referenced use different library and I wasn't able to replicate the behaviour with what's being used in the app. Let me know if the current state is acceptable.

parrot9design:
>Yearly queries were often failing. It's now replaced with the `Max` timeframe which works quite good, and uses monthly candles instead.

Let's update the label to `All time`. Noticed afSUI doesn't play so well with this timeframe, but it matches what's on Coingecko so not sure if we can improve it. Otherwise it looks good. 

---

>>Some of the values on the y axis aren't being displayed properly. The line chart often shows 0-1 value along the side (particularly for stablecoins), and the price shown on hover often doesn't provide enough precision – based on the padding, I think it's being cut off.
>
>This was so fkin painful 😅
>Came up with showing the highlighted current price which helps with spacing out the Y axis ticks but messes up the transition animations of the charts sometime. Let me know what feels better to you, can be compared with this deployment:
>https://balanced-network-interface-dui76wtz5-balanced-defi.vercel.app/

Sorry about that. 😅 I prefer the version with the current price highlighted, so at least your pain wasn't for nothing. 😂 

---

>>Trying to reveal the price at a specific point on the chart is quite difficult to do on a touch screen. You have to tap very intentionally, and you can't move along the line/candles like you can with a mouse hover on desktop. Not sure how much flexibility you have with the chart infrastructure, but is it possible to make it more seamless to reveal the data like the charts on the Stats page?
>
>Still having issues with this. The charts you referenced use different library and I wasn't able to replicate the behaviour with what's being used in the app. Let me know if the current state is acceptable.

That's all good. I wasn't sure if this library would give us as much flexibility, but wanted to run it past you just in case we could improve it. Happy to ship without this behaviour in place (having the current price highlighted already helps some), perhaps something we can look to enhance with v2.

Just one thing: are we able to prevent the chart from moving on mobile?

https://github.com/user-attachments/assets/46776dc2-e950-43fc-beab-fed2f2cff9a1

---

Just posting this on the off-chance we have a better data source: the all-time chart for bnUSD sucks. It looks like it really tanked between March 2023 – July 2024, even though it didn't. I remember having to jump countless hurdles with Coingecko to get it to reflect the actual price, didn't expect it to burn us in our own interface, as well. 😅😭
Screenshot 2025-10-07 at 10 43 50 AM

PR #1760, opened 2026-04-09
Update migration page

hetfly:


parrot9design:
Looks good. Just confirming with the SODAX team that the current values should still reflect your initial migration amount.

Also wondering if we should keep the xSODA and current value in place during the unstaking process, so that line is only missing for SODA that hasn't been staked or is fully unstaked.
Screenshot 2026-04-10 at 10 06 23 AM

hetfly:
>Also wondering if we should keep the xSODA and current value in place during the unstaking process


Both are 0 when unstaking

parrot9design:
The SODAX team confirmed that the values should increase very slowly, so I think this is all working as expected.

Let's merge on Monday.

parrot9design:
Some of my SODA has completed the lock-up time. Is it expected that it will still appear under the Migrate tab with the ability to stake it? Apparently it's not staked, yet it still has an xSODA value attached.
Screenshot 2026-04-13 at 11 51 41 AM

And if I click the Stake button, I see this modal:
Screenshot 2026-04-13 at 11 53 35 AM

I do have some SODA in my wallet, but it's unclear if it came from this.

Here's my wallet address: 0x9c7d2E8efA7e6885e90C0CFb8D1A8fa3dad57538

hetfly:
@parrot9design looked at the sonic tracker, nothing indicates that you received the migrated tokens automatically.

There is a method in the sdk that should claim the tokens, I've implemented it but I can't test it myself. My earliest unlock is on 27th. Have a look and let me know if it works please.

Also, the xSoda row should be hidden for all but staked migration items now.

parrot9design:
Doesn't look like anything has changed. I can still see the xSODA balance and Stake button for the SODA that is now available, and there's no option to claim. 🤔

parrot9design:
For migrations that are currently unstaking, it now shows the unstaking date twice. The top one should always be the unlock date, which won't always be the same.
Screenshot 2026-04-17 at 10 43 23 AM

When I tried to test staking/unstaking with other migrations, I got these errors:
Screenshot 2026-04-17 at 10 54 12 AMScreenshot 2026-04-17 at 10 54 20 AM

I think for some, especially unstaking, it's because the unstaking transaction went through but the UI didn't update, so then when I try to unstake again I get the error.
Screenshot 2026-04-17 at 11 05 03 AMScreenshot 2026-04-17 at 11 05 55 AM

Screenshot 2026-04-17 at 11 06 53 AM Screenshot 2026-04-17 at 11 06 19 AM



---

For the modals, let's remove the message for SODA, so it just looks like this:

Screenshot 2026-04-17 at 10 46 23 AM

For xSODA, we could update the message to this:
Screenshot 2026-04-17 at 10 50 24 AM

And if unstaking has begun, we could use this:
Screenshot 2026-04-17 at 11 02 39 AM

---

I claimed this migration twice, but it still shows in the UI and appears that it will let me continue to claim it as many times as I like.  
Screenshot 2026-04-17 at 11 11 24 AM

hetfly:
Lol, apologies for this mess @parrot9design 

main issue was that the transactions are using migration item index. Once I tested the claiming, the item got filtered out. That shifted the indexing in the app and broke all the other transactions - including the claim you were trying. You couldn't claim it multiple times since the tx was using wrong index, that's the reason it didn't disappear from the UI.

parrot9design:
This looks much better, thank you!

parrot9design:
Tried testing a few more transactions that had unlocked. I was able to claim my xSODA without any issues, but the SODA claims all returned these errors:

Screenshot 2026-04-22 at 10 57 41 AM
Screenshot 2026-04-22 at 10 57 49 AM

